PR #1, opened 2023-10-08
tests

frnandu:
initial tests with a basic mock relay implementation for responding with NIP65


lib/bip340.dart
leo-lox, line 1: I think we schould type this, also maybe call it publicKeyHex, publicKeyBech instead of publicKeyHr

lib/event.dart
leo-lox, line 1: call it NostrEvent? so its clear that this is the dart representation of a raw Nostr event.
The idea is that we can exted it to to nostr event we are using inside the lib

lib/bip340.dart
frnandu: Do you mean type this from zero? Why? What's the current problem with this code? I can only imagine it will be needed an interface to allow for client native implementations of signing/verifying since on android the acinq lib is much faster. And for web probalby also.

lib/event.dart
frnandu: Can you explain it a bit more? What will the extended one have in addition to this?

lib/nips/Nip65.dart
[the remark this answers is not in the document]
frnandu: True. I copied the code. We can clean it up. Do you have a cleaner version from camelus? Feel free to replace it.

lib/event.dart
leo-lox: for example if we decide to do verifing we can add/extend with a bool verified. Its basically just an abstraction so its clear what the "pure" nostr event is and what we added

lib/nips/Nip65.dart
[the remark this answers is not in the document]
leo-lox: i tried to clean it up

lib/bip340.dart
frnandu: I'll remove them

PR #2, opened 2023-10-10
ranking algo

leo-lox:


leo-lox:
we need to discuss what readOnly and writeOnly means, schould read/write be included if we are looking for readOnly?
@see `relay_ranking_test.dart l:365`

frnandu:
> we need to discuss what readOnly and writeOnly means, schould read/write be included if we are looking for readOnly? @see `relay_ranking_test.dart l:365`

if you want to check marker for readOnly just do marker == ReadWriteMarker.readOnly ?

PR #33, opened 2024-11-01
close in requests usecase

leo-lox:


frnandu:
Should we be adding more specific methods to the engine, instead of trying to unify as much as possible?

leo-lox:
agree, how do you see the closeSub() in requests usecase?
Right now because of the diffrent global connection it needs to be seperate. As soon as I unify the relay and relayJit impl close sub can be unified.

PR #37, opened 2024-11-09
check cache results against limit

leo-lox:
fixes a bug on req when cache is partly filled but does not satisfy the full limit param.
=> Dont delete authors and send to network in this case


frnandu (approved):

frnandu:
could you add some test code for this?

leo-lox:
sure just wanted to check first if thats the right approach

PR #38, opened 2024-11-11
make filter immutable

leo-lox:
fixes a bug where ndk changes lists of the lib user


frnandu:
would be good to have test code

PR #46, opened 2024-11-16
nip05 usecase

leo-lox:
would be nice to have a get nip05 by pubkey in this => return the valid or invalid nip05 [check metadata first, check db for valid]
Could be in one class, current VerifyNip05 is very explicit


lib/domain_layer/usecases/nip05/verify_nip_05.dart
```diff
@@ -0,0 +1,80 @@
...
+  ///
+  /// [nip05] the nip05 identifier
+  /// [pubkey] the public key
+  /// returns the [Nip05] object
+  Future<Nip05> check({required String nip05, required String pubkey}) async {
+    if (nip05.isEmpty || pubkey.isEmpty) {
+      throw Exception("nip05 or pubkey empty");
+    }
+
+    final result = Nip05(pubKey: pubkey, nip05: nip05);
```
frnandu, line 35: Can this be moved down just before it is needed?

lib/domain_layer/usecases/nip05/verify_nip_05.dart
```diff
@@ -0,0 +1,80 @@
...
+  Future<Nip05> _performCheck(String nip05, String pubkey, Nip05 result) async {
+    Nip05? networkResult;
+    try {
+      networkResult = await _nip05Repository.requestNip05(nip05, pubkey);
+    } catch (e) {
+      networkResult = null;
+    }
+
+    if (networkResult != null) {
+      result.valid = networkResult.valid;
```
frnandu, line 74: Shouldn't it also copy the `relays` and `networkFetchTime` properties?

lib/data_layer/models/nip_05_model.dart
```diff
@@ -0,0 +1,13 @@
+import 'package:ndk/domain_layer/entities/nip_05.dart';
+
+/// model for converting json <==> dart object
+class Nip05Model extends Nip05 {
```
frnandu, line 4: Why is this necessary?

lib/domain_layer/usecases/nip05/verify_nip_05.dart
```diff
@@ -0,0 +1,80 @@
...
+      _inFlightRequests.remove(nip05);
+    }
+  }
+
+  Future<Nip05> _performCheck(String nip05, String pubkey, Nip05 result) async {
+    Nip05? networkResult;
+    try {
+      networkResult = await _nip05Repository.requestNip05(nip05, pubkey);
+    } catch (e) {
+      networkResult = null;
```
frnandu, line 70: Doesn't this cause the result to be cached without a timestamp, causing it to everytime go try to fetch from network again?
If it gets an exception because of 404 or other shit, we should also set `result.networkFetchTime` to current timestamp, otherwise it will non-stop try to fetch it.

frnandu (changes requested):

lib/domain_layer/usecases/nip05/verify_nip_05.dart
leo-lox: its directly accessed afterwards, just a check that pubkey or nip05 is not  ""

lib/domain_layer/usecases/nip05/verify_nip_05.dart
leo-lox: good catch, added tests

lib/domain_layer/usecases/nip05/verify_nip_05.dart
leo-lox: good catch, added tests

lib/data_layer/models/nip_05_model.dart
leo-lox: we can remove this for now. Its could be used to convert from and to json. But we are not using it right now. 
Maybe this canges if we save in db so I would like to leave it for now

lib/domain_layer/usecases/nip05/verify_nip_05.dart
frnandu: I had in mind the `result` var which is only needed in 
`final request = _performCheck(nip05, pubkey, result);`
much lower...
it could even be removed if passing directly a new instance of Nip05 into the `_performCheck` method....

lib/data_layer/models/nip_05_model.dart
frnandu: Can't we have `.fromJson` and `.toJson` in the `Nip05` class?
Analogously to for example `Metadata.fromJson` ?

leo-lox: yes its just cleaner, no dependencies in the entity

lib/domain_layer/usecases/nip05/verify_nip_05.dart
leo-lox: I see, removed it

frnandu:
Nice job!

leo-lox:
good catch on the review!

PR #51, opened 2024-11-25
Melos

### LINKED ISSUE #24: melos

frnandu:
Split into monorepos, so that ndk can dependency free from flutter
- ndk
- amber
- rust_verifier
- objectbox
- sample-app

fixes #24


packages/rust_verifier/pubspec.yaml
```diff
@@ -0,0 +1,50 @@
...
+  collection: ^1.18.0
+  path_provider: ^2.1.4
+  convert: ^3.1.1
+  pointycastle: ^3.9.1
+  plugin_platform_interface: ^2.1.7
+  logger: ^2.4.0
+  rxdart: ^0.28.0
+  flutter_rust_bridge: 2.6.0
+  rust_lib_ndk: ^0.1.3
+  ndk:
```
leo-lox, line 36: does this work when publishing a package? Is melos smart enouth to replace with the new deployed version when pushing?

.github/workflows/tests.yaml
```diff
@@ -3,13 +3,23 @@ on: [push, workflow_dispatch]
...
       - uses: actions/checkout@v1
-      - uses: subosito/flutter-action@v1
-      - run: flutter doctor
-      - run: flutter test --coverage
-      - run: flutter test ./example
+      - uses: dart-lang/setup-dart@v1
+      - run: dart pub get
+      - run: dart format --output=none .
+      - run: dart analyze --no-fatal-warnings
+      - run: dart test --coverage="coverage"
```
leo-lox, line 15: can we use `melos run test:dart/flutter` here?
Right now there seems to be a dependency issue on rust_lib_ndk

packages/ndk/pubspec.yaml
```diff
@@ -0,0 +1,35 @@
...
+  convert: ^3.1.1
+  pointycastle: ^3.9.1
+  logger: ^2.4.0
+  rxdart: ^0.28.0
+  isar: ^4.0.0-dev.14
+
+dev_dependencies:
+  build_runner: ^2.4.11
+  mockito: ^5.0.17
+  objectbox_generator: any
```
leo-lox, line 34: possible to remove?

leo-lox (approved):
nice work!

.github/workflows/tests.yaml
frnandu: for now it is running tests only for ndk because of 
`working-directory: ./packages/ndk`
but to be decided how to run that on all packages once we HAVE tests for other packages besides ndk only

packages/rust_verifier/pubspec.yaml
frnandu: not sure, to be tested/researched

packages/ndk/pubspec.yaml
frnandu: objectbox_generator yes,
mockito is needed for sure,
build_runner I'm not sure, it is used for generating isar when changing something in the db models

.github/workflows/tests.yaml
leo-lox: I created a ticket

PR #53, opened 2024-11-28
Lists usecase public use

leo-lox:
not sure why `if (name == null || event.kind != Nip51List.RELAY_SET) ` was in lib/domain_layer/entities/nip_51_list.dart (removed it so other lists work)
Is there a reason for it @frnandu ?


leo-lox:
removed the default kind.
I also plan to refactor in a seperate branch to use broadcast usecase, and there we can clean up more and also add tests for other types of lists

frnandu (approved):

PR #54, opened 2024-11-28
fix: Missing events in concurrency check when stream attaches

leo-lox:


leo-lox:
case:
one request r0 with filter f0 is in flight and not fully processed another request r1 with filter f0 gets added. 

This fix makes sure that r1 gets the same events as r0 even when attached at a later point in time

leo-lox:
I set it to 2000 should be plenty enough and prevent memory leaks

PR #72, opened 2024-12-12
await for broadcast cases

leo-lox:
added await in remaingin usecases where broadcast is used


leo-lox:
we might reafactor in the future to use streams instead (in the usecases)

PR #76, opened 2024-12-17
nip-57 Zaps

### LINKED ISSUE #75: nip57 zap support

frnandu:
fixes #75


packages/ndk/example/zaps/zap.dart
```diff
@@ -0,0 +1,53 @@
+// ignore_for_file: avoid_print
+
+import 'dart:io';
+
+import 'package:ndk/domain_layer/usecases/zaps/zap_receipt.dart';
+import 'package:ndk/domain_layer/usecases/zaps/zaps.dart';
```
leo-lox, line 6: can you add zaps to the default exports

packages/ndk/lib/domain_layer/usecases/lnurl/lnurl.dart
```diff
@@ -0,0 +1,98 @@
+import 'dart:convert';
+
+import 'package:http/http.dart' as http;
+
+import '../../../shared/logger/logger.dart';
+import 'lnurl_response.dart';
+
+/// LN URL utilities
+abstract class Lnurl {
```
leo-lox, line 9: why is this abstract? 
I see that all methods are static but maybe you had something in mind?

leo-lox (approved):
strong!
Just need to fix some imports.
And in general (future thinking) is it possible to switch the underlying asset like ecash?

leo-lox:
http client is not injected

packages/ndk/lib/domain_layer/usecases/lnurl/lnurl.dart
frnandu: Just that it doesn't keep state, just utilities methods so far. Can that still be a usecase? Not sure if later it will....

frnandu: you're right in that it should not be abstract and keep the http transport somehow inside

PR #78, opened 2024-12-20
web_socket_client transport impl + quick fixes for reconnect

frnandu:


leo-lox (approved):
lets test in real conditions and then make onReconnect required

PR #80, opened 2024-12-21
test with dead relay

leo-lox:


leo-lox:
the out controller is no longer closed in relay manager. 
Instead its in requests and relay manager closes the network stream `_checkNetworkClose()`
Not ideal because we need to check if the relay is still online.

Also changed the DEFAULT_WEB SOCKET_CONNECT_TIMEOUT to 4s so ndk startup is fast enough even if some relays are down

frnandu:
Important and much needed tests, nice job!

frnandu (approved):

frnandu (changes requested):

PR #89, opened 2024-12-26
Cleanup

franzaps:
Basic NDK package cleanup, with the goal of making it more clear to understand, add TODOs and remove files that do not belong.

## Filesystem

 - Remove ios folder - why was it there? This no longer depends on Flutter
 - Remove Isar - should not belong to the core ndk
 - Renamed bad file names with all uppercase words

## Code warnings

 - Removed `public_member_api_docs` from the linter, at the very least until there is interest to actually perform the task (over 500 warnings gone)
 - Refactored string interpolations
 - Refactored all enums for proper casing
 - Constants start with lowercase k in Dart, refactored all those
 - Left `TODO`s throughout the codebase but clarified several of them, they should be kept actionable and descriptive and if it's an important issue moved to an issue tracker

## Comment on variable modifiers and names

Types in Dart are inferred, not only using them on the left hand of the assignment makes them redundant (e.g. `ZapReceipt receipt = ZapReceipt.fromEvent(event)` looks mega redundant) but it discourages thinking about proper descriptive variable names:

`List<NwcConnection> list = something.toList();`

Should in my opinion be:

`final nwcConnections = something.toList();`

Now every time you see `nwcConnections` it's clearer than `list`. While developing you can anyways always hover over to get the type.

On top of that, it sometimes causes to forget the final modifier. Always prefer finals, this helps reduce mutation bugs which are common.

`String test = "mystring"` -> `final testString = "mystring"`

## Other minor questions

 - Is it `readme.md` or `README.md` - what is the convention?
 - Not right to have static ints, it creates issues when subclassing for example: `static const kZapRequestKind = 9734;` can't be called `kKind` as it shadows that of `Nip01Event`
 - Can we name directories just `domain` vs `domain_layer`? it is sort of obvious that directories are used as layers


packages/ndk/analysis_options.yaml
```diff
@@ -16,4 +16,3 @@ linter:
   # `// ignore_for_file: name_of_lint` syntax on the line or in the file
   # producing the lint.
   rules:
-    - public_member_api_docs
```
leo-lox, line 19: before we had not linter for api_docs. So I am not so sure if removing is the best way forward. Do you know if there is a warning level below info?
Ideally we have something that still screams: hey you forgot api docs here

packages/ndk/example/basic_test.dart
```diff
@@ -29,7 +29,7 @@ void main() async {
             '3bf0c63fcb93463407af97a5e5ee64fa883d107ef9e558472c4eb9aaaefa459d'
           ],
           // Query for text note events (kind 1)
-          kinds: [Nip01Event.TEXT_NODE_KIND],
+          kinds: [Nip01Event.kKind],
```
leo-lox, line 32: what is the reason for kKind? Dont quite understand

If its for generic kind getter then we schould enforce it via abstract class, but not sure if I understand

packages/ndk/lib/domain_layer/usecases/relay_manager.dart
```diff
@@ -43,7 +43,7 @@ class RelayManager<T> {
   final EngineAdditionalDataFactory? engineAdditionalDataFactory;
 
   /// Are reconnects allowed when a connection drops?
-  bool _allowReconnectRelays = true;
+  bool allowReconnectRelays = true;
```
leo-lox, line 46: why change to public?

packages/ndk/lib/domain_layer/usecases/zaps/zap_request.dart
```diff
@@ -2,10 +2,10 @@ import '../../entities/nip_01_event.dart';
 
 /// Zap Request
 class ZapRequest extends Nip01Event {
-  static const int KIND = 9734;
+  static const kZapRequestKind = 9734;
```
leo-lox, line 5: maybe we need to rethink the kind naming:
kindZapRequestKind is dublicated
Proposal:

either keep short kTextNote or kindTextNote everywhere, also in definition class =>
Nip01Event.kTextNote
Nip01Event.kindTextNote

instead of  
Nip01Event.kKind

leo-lox (commented):
Nice work, thx for working on this.
Two things that I noticed

1.)
We need to fix our naming with kinds (thx for pointing this out) and include it in our docu.

2.)
where did the isar stuff go? 
Am I missing something?

frnandu:
> 2.)
> where did the isar stuff go? 
> Am I missing something?
> 

I'll put the isar stuff on a separate monorepo

leo-lox:
@franzaps 
> Is it readme.md or README.md - what is the convention?

I will add a contributing guide soon (=>README.md)

>Can we name directories just domain vs domain_layer? it is sort of obvious that directories are used as layers

I think its obvious for people that have already worked with layered architecture and to keep it that way I find the reminder helpful


> Comment on variable modifiers and names

While writing code I like to have the type upfront so when I edit around, I can't break stuff in other places, but I guess it comes down to personal preferences.

packages/ndk/lib/domain_layer/usecases/relay_manager.dart
frnandu: I had getter/setter which I use in Yana when going to background/foreground, but @franzaps is right that it can simple be public instead.

packages/ndk/example/basic_test.dart
frnandu: should be Nip01Event.kTextNodeKind

packages/ndk/analysis_options.yaml
frnandu: Maybe we can increase the severity of other linter rules, and leave this one as info (there is no level below info AFAIK) and hopefully we'll add docs to everything in time.

packages/ndk/example/basic_test.dart
frnandu: @franzaps can you change that? 
We can merge after that.

frnandu:
can you fix the conflict in the branch @franzaps so we can merge

packages/ndk/analysis_options.yaml
franzaps: I feel this linter rule goes way beyond what makes sense to document. It's literally showing warnings on hundreds of methods. Maybe it's better to start writing on the most important methods. Also general documentation would be super helpful. 

As my 2c feedback, feel free to ignore, I find it a bit hard to understand the project: first it's a *lot* of code, then when I browse the code I can't find many patterns, also I need to go study clean architecture. Maybe a general documentation site would add a lot of value for newbs like me

packages/ndk/example/basic_test.dart
franzaps: In a class you would use one member to refer to something, i.e. `kind`, and then override in each subclass. You then know that no matter where you are, you call `kind` and you get the correct value.

I understand that kinds in this context need to be accessed without instantiating a class - so statics are a way of doing it. Calling them exactly the same everywhere (`kKind`) was my attempt to normalize it. But then we run into the issue of clashing naming in subclasses with statics. I'll revert to `kTextNodeKind` for this reason. I prefer the approach I took in purplebase for this (a map)

franzaps:
> > Comment on variable modifiers and names
> 
> While writing code I like to have the type upfront so when I edit around, I can't break stuff in other places, but I guess it comes down to personal preferences.

You won't break stuff in other places when you edit around, the compiler won't let you. 

Personal preferences - sure. What are your thoughts on my specific points? For instance that there are a lot of unnecessary non-final variables.

@fmar thanks for moving the Isar stuff! Did you manage to do that?

franzaps:
Is this good or are further changes needed?

frnandu:
> @fmar thanks for moving the Isar stuff! Did you manage to do that?
Will do in a new PR

PR #91, opened 2024-12-28
Handle canceling of streamSubscriptions

frnandu:


packages/ndk/lib/domain_layer/usecases/zaps/zaps.dart
```diff
@@ -141,11 +143,14 @@ class Zaps {
...
                 .closeSubscription(zapResponse.zapReceiptResponse!.requestId);
+            if (zapResponse.streamSubscription!=null) {
+              zapResponse.streamSubscription!.cancel();
+            }
             Logger.log
                 .w("timed out waiting for zap receipt for invoice $invoice");
           });
 
-          zapResponse.zapReceiptResponse!.stream.listen((event) {
+          zapResponse.streamSubscription = zapResponse.zapReceiptResponse!.stream.listen((event) {
```
leo-lox, line 153: why is the stream sub on the zapResponse?
As far as I can work it out its only used in zaps.dart, would be good to avoid side effects

leo-lox:
do you know if there is a linter rule to enforce closing subs?

frnandu:
> do you know if there is a linter rule to enforce closing subs?

https://dart.dev/tools/linter-rules/cancel_subscriptions ?

PR #121, opened 2025-02-25
fix: blossom parsing issues

### LINKED ISSUE #120: Blossom blob upload fails
Blossom uploads are failing with error `type 'String' is not a subtype of type 'int?'` 

This is because `BlobNip94` has values which are not strings, all values must be strings

https://github.com/relaystr/ndk/blob/8897201eafa0d5de43a3686796943ddfb9c971e8/packages/ndk/lib/domain_layer/entities/blossom_blobs.dart#L172

1-leo:
fixes #120

### DIFF
--- a/packages/ndk/lib/domain_layer/entities/blossom_blobs.dart
+++ b/packages/ndk/lib/domain_layer/entities/blossom_blobs.dart
@@ -1,5 +1,15 @@
 import 'dart:typed_data';
 
+_parseSize(dynamic size) {
+  if (size is int) {
+    return size;
+  } else if (size is String) {
+    return int.tryParse(size) ?? 0;
+  } else {
+    return 0;
+  }
+}
+
 /// Descriptor of a blob - when getting a blob from a server
 class BlobDescriptor {
   /// server url e.g. https://example.com/&lt;sha256&gt;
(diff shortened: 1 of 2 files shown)


packages/ndk/lib/domain_layer/entities/blossom_blobs.dart
```diff
@@ -169,7 +179,8 @@ class BlobNip94 {
       mimeType: json['m'] ?? '',
       sha256: json['x'] ?? '',
       originalsha256: json['ox'] ?? '',
-      size: json['size'] ?? 0,
+      // parse int from string
+      size: _parseSize(json['size']),
       dimenssions: json['dim'],
```
v0l, line 184: You need to test this too the format is a string `"1920x1080"`  it wont be parsed into `int?`

leo-lox: according to the specs its used for file byte size, is this used differently in implementations that exists?

v0l: No i mean `dim` tag not `size`

leo-lox: got it

packages/ndk/test/entities/blob_nip94_test.dart
```diff
@@ -55,5 +55,36 @@ void main() async {
...
+      };
+
+      final Map<String, dynamic> malformedJson2 = {
+        "url": '',
+        "sha256": '',
+        "size": "512",
+        "dim": 100,
+        "type": "notype",
+        "nip94": {
+          "dim": 100,
```
v0l, line 79: Why handle this? `dim 100` is invalid

v0l: Ok i thought you added this because one server IS responding like this

PR #124, opened 2025-02-26
fix: blossom nip94 parsing List<String>?

v0l:
BUD-08 is flawed:
Ref: https://github.com/hzrd149/blossom/pull/60

### DIFF
--- a/packages/ndk/lib/domain_layer/entities/blossom_blobs.dart
+++ b/packages/ndk/lib/domain_layer/entities/blossom_blobs.dart
@@ -114,9 +114,6 @@ class BlobNip94 {
   /// SHA-256 hexencoded string of the file
   final String sha256;
 
-  /// SHA-256 hexencoded string of the original file
-  final String originalsha256;
-
   /// size of file in bytes
   final int? size;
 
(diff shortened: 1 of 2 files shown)


packages/ndk/lib/domain_layer/entities/blossom_blobs.dart
```diff
@@ -178,18 +174,17 @@ class BlobNip94 {
       url: json['url'] ?? '',
       mimeType: json['m'] ?? '',
       sha256: json['x'] ?? '',
-      originalsha256: json['ox'] ?? '',
```
leo-lox, line 181: why remove original hash? Is this no longer up to date? 
(listed in https://github.com/nostr-protocol/nips/blob/master/94.md)

leo-lox (commented):
Thx for the pr!
should we update the code to work with arrays after the merge?
https://github.com/hzrd149/blossom/pull/60
Bit confused here

packages/ndk/lib/domain_layer/entities/blossom_blobs.dart
v0l: `ox` is not used in blossom, though it is listed as NIP94 field, `ox` is used for NIP-96 only

v0l:
> Bit confused here

Well originally i just wanted to fix the bug, but then decided to update the spec instead, ill make another PR if the spec changes.

packages/ndk/lib/domain_layer/entities/blossom_blobs.dart
leo-lox: got it

PR #128, opened 2025-03-03
Broadcast stop early

1-leo:
- adds timeouts to broadcast
- adds percent consider broadcast done

`=> considerDonePercent || timeout`


frnandu (approved):

leo-lox:
I moved the settings to ndkConfig, still possible to override when using the broadcast service directly. 
I think this is more useful otherwise we need to pass the arg around in each use case

PR #135, opened 2025-03-14
feat: add a tag to zaps

PatrickGeyer:


PatrickGeyer:
Btw, not sure about removing pubkey query param, this might break validation since anyone can publish a ZapReceipt referencing just the a event!

frnandu:
Thanks!

PR #142, opened 2025-03-26
adds close() to dbs

1-leo:
adds a close() to the cache manager, used to enable closing the database.
Needed to close db in a bg thread

adds attach to objectbox constructor


frnandu (approved):
looks good, thanks

PR #143, opened 2025-03-27
getTags not static

1-leo:
@frnandu was there a reason why getTags() static? (only used once in nwc)


frnandu (approved):
no, should be good not static, thanks

PR #164, opened 2025-05-26
Fix connection to a bad relay blocking event delivery

### LINKED ISSUE #154: RELAY_SETS engine takes 5s to load some events
visible in https://github.com/nostrlabs-io/zap-stream-flutter

frnandu:
Make reconnect to relays async to not block other relays responding immediately in RELAYS_SET engine

fixes #154 

@v0l try this on zapstream, should fix the 5s delay that `wss://relay.nostr.bg` bad relay was causing.


v0l:
YES its much faster!

v0l:
There also seems to be a similar issue with `relay.nostr.bg` when publishing

PR #188, opened 2025-07-09
Add nip 07 support

nogringo:
Everything is in the title.

Info: since getPublicKey is async in web extensions, I added this method : `await Nip07EventSigner().getPublicKeyAsync();`


leo-lox (approved):
Thank You! 
I will add it to the documentation, therefore merging it into  `docs-nip07-event-signer`

PR #190, opened 2025-07-12
Add sembast cache manager

nogringo:
A new cache manager based on sembast https://pub.dev/packages/sembast

### DIFF
--- a/packages/sembast_cache_manager/example/sembast_cache_manager_example.dart
+++ b/packages/sembast_cache_manager/example/sembast_cache_manager_example.dart
@@ -1,0 +1,161 @@
+import 'dart:io';
+import 'package:ndk/domain_layer/entities/nip_05.dart';
+import 'package:ndk/ndk.dart';
+import 'package:sembast/sembast_io.dart';
+import 'package:sembast_cache_manager/sembast_cache_manager.dart';
+
+Future<void> main() async {
+  print('🚀 SembastCacheManager Example\n');
+
+  // 1. Create a temporary database
+  final tempDir = await Directory.systemTemp.createTemp('sembast_cache_example_');
+  final dbPath = '${tempDir.path}/cache.db';
+  print('📁 Database path: $dbPath');
+
+  // 2. Open the Sembast database
+  final database = await databaseFactoryIo.openDatabase(dbPath);
+  print('✅ Database opened successfully');
+
+  // 3. Create the cache manager
+  final cacheManager = SembastCacheManager(database);
+  print('✅ Cache manager created\n');
+
+  try {
+    // 4. Create some sample data
+    print('📝 Creating sample data...');
+    
+    // Create a sample event
+    final event = Nip01Event(
+      pubKey: 'npub1abc123def456ghi789jkl012mno345pqr678stu901vwx234yz567890',
+      kind: 1, // Text note
+      tags: [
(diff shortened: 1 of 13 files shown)


packages/sembast_cache_manager/pubspec.yaml
```diff
@@ -0,0 +1,16 @@
+name: sembast_cache_manager
+description: A Sembast-based cache manager for the NDK (Nostr Development Kit) library, providing persistent storage for Nostr protocol data.
+version: 1.0.0
+homepage: https://github.com/nogringo/sembast-cache-manager
```
leo-lox, line 4: change these to the ndk link

packages/sembast_cache_manager/example/sembast_cache_manager_example.dart
```diff
@@ -0,0 +1,161 @@
...
+import 'package:ndk/domain_layer/entities/nip_05.dart';
+import 'package:ndk/ndk.dart';
+import 'package:sembast/sembast_io.dart';
+import 'package:sembast_cache_manager/sembast_cache_manager.dart';
+
+Future<void> main() async {
+  print('🚀 SembastCacheManager Example\n');
+
+  // 1. Create a temporary database
+  final tempDir = await Directory.systemTemp.createTemp('sembast_cache_example_');
```
leo-lox, line 11: can the init step be moved to the constructor of SembastCacheManager, maybe with params for the db path.
Similar to https://github.com/relaystr/ndk/blob/master/packages/objectbox/lib/data_layer/db/object_box/db_object_box.dart makes it more convenient to use

leo-lox: and adjust readme etc., accordingly

leo-lox (changes requested):
Looks solid. Thanks for the PR!
Can you move the init code with defaults to the constructor?
Makes it more convenient to use

packages/sembast_cache_manager/example/sembast_cache_manager_example.dart
nogringo: It's better to let the developper to create his sembast DB, maybe he want encryption, maybe he do not want to support flutter web. We can provide them a simple function to simply create their cache manager.

```dart
import 'dart:io';

import 'package:flutter/foundation.dart';
import 'package:path_provider/path_provider.dart';
import 'package:sembast_web/sembast_web.dart';
import 'package:sembast/sembast_io.dart';
import 'package:path/path.dart' as p;

Future<Database> getDatabase([String dbName = "ndk_cache_manager"]) async {
  if (kIsWeb) {
    var factory = databaseFactoryWeb;
    return await factory.openDatabase('test');
  }

  final Directory appDocumentsDir = await getApplicationDocumentsDirectory();
  return await databaseFactoryIo.openDatabase(
    p.join(appDocumentsDir.path, '$dbName.db'),
  );
}

void main() async {
  final cacheManager = SembastCacheManager(await getDatabase());
}
```

leo-lox: Yeah, this also works. Or with a named constructor.
I think it should be easy for devs to get started that don't know about sembast (like me) while still providing all the options.

leo-lox (approved):
nice work!

PR #203, opened 2025-07-24
fix parsing rumor event with no sig

### LINKED ISSUE #202: Gift wrap Nip01Event parsing err
Parsing the rumor fails, probably because 'sig' is not set

Image

1-leo:


nogringo (approved):
Perfect !

PR #207, opened 2025-08-02
Nip46 event signer

nogringo:


packages/nip46_event_signer/lib/src/nip46_event_signer.dart
```diff
@@ -0,0 +1,240 @@
...
+    final decryptedText = await remoteRequest(request: request);
+    return decryptedText;
+  }
+
+  @override
+  Future<String?> decryptNip44({
+    required String ciphertext,
+    required String senderPubKey,
+  }) async {
+    final request = {
```
leo-lox, line 139: request could be typed

leo-lox (changes requested):
Interesting approach using a separate NDK instance. Actually has some benefits. 
Have you thought about integrating it directly? Not possible?

I think it makes sense to move the code into the NDK package. We can still use a separate NDK if it makes sense architecturally.

I marked one problem with `generateRandomString()`.

Thx for implementing!

leo-lox:
When merging, we should also add a method to the accounts usecase to make it even more integrated.

packages/nip46_event_signer/test/nip46_event_signer_test.dart
```diff
@@ -0,0 +1,184 @@
...
+    late Nip46EventSigner signer;
+    late ConnectionSettings connectionSettings;
+
+    setUp(() {
+      connectionSettings = ConnectionSettings(
+        privateKey:
+            "7a8317f947fff0526749e9fe53f79def8eb0afd378c01058f37140cc8732fecc",
+        remotePubkey:
+            "b836aab8e635e41e62b832d68dc0f1857b717689df248290be9efa02f02de672",
+        relays: ["wss://relay.nsec.app"],
```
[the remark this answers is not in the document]
nogringo: yes we can

packages/nip46_event_signer/lib/src/nip46_event_signer.dart
nogringo: What do you mean ?

1-leo: A class with to/fromJson

packages/ndk/lib/data_layer/repositories/signers/nip46_event_signer/nip46_event_signer.dart
```diff
@@ -0,0 +1,231 @@
...
+  Ndk? ndk;
+  NdkResponse? subscription;
+
+  final _pendingRequests = <String, Completer<dynamic>>{};
+
+  String? _cachedPublicKey;
+
+  late Bip340EventSigner localEventSigner;
+
+  Nip46EventSigner({required this.connectionSettings}) {
```
1-leo, line 28: a good thing to have would be to pass in other EventVerifiers (like rust verifier) as an optional parameter

1-leo (approved):
Noticed some minor issues. Overall, very solid!
We can fine-tune the API in another PR if you need this feaure soon.
On Wednesday we should discuss whether to make getPublicKeyAsync the default.

frnandu (changes requested):
see https://github.com/relaystr/ndk/pull/213

frnandu (approved):

1-leo:
- [ ] tests (needs support in mock relay)
- [ ] docs

PR #213, opened 2025-08-06
nip46 bunkers usecase

frnandu:


packages/ndk/lib/domain_layer/usecases/bunkers/bunkers.dart
```diff
@@ -0,0 +1,182 @@
...
+        senderPubKey: remotePubkey,
+      );
+
+      final response = jsonDecode(decryptedContent!);
+
+      if (response["id"] != request.id) continue;
+
+      if (response["result"] == "auth_url") {
+        // _streamController.add(AuthRequired(response["error"]));
+        // TODO what can we do about this?
```
nogringo, line 104: Normaly the app must inform the user that he have an authentication to do.
https://github.com/nostr-protocol/nips/blob/master/46.md#auth-challenges

frnandu: let's maybe add an optional authCallback  param function that user can pass to handle it ?

frnandu: but the signer also needs this right? for when trying to perform operations that were not in the list of permissions when initially connected?

packages/ndk/example/bunkers_example.dart
```diff
@@ -0,0 +1,32 @@
+import 'package:ndk/ndk.dart';
+
+Future<void> main() async {
+  final ndk = Ndk.emptyBootstrapRelaysConfig();
+
+  try {
+    await ndk.accounts.loginWithBunkerUrl(
+        bunkerUrl:
+            "bunker://a1fe3664f7a2b24db97e5b63869e8011c947f9abd8c03f98befafd27c38467d2?relay=wss://relay.nsec.app&secret=devsecret123",
+        bunkers: ndk.bunkers);
```
1-leo, line 10: can we pass in the bunkers usecase by default?

packages/ndk/lib/domain_layer/usecases/bunkers/bunkers.dart
```diff
@@ -0,0 +1,189 @@
...
+  Bunkers({
+    required Broadcast broadcast,
+    required Requests requests,
+  })  : _broadcast = broadcast,
+        _requests = requests;
+
+  /// Connects to a bunker using a bunker URL (bunker://)
+  Future<BunkerConnection?> connectWithBunkerUrl(
+    String bunkerUrl, {
+    Function(String)? authCallback,
```
1-leo, line 32: a short explanation for what the auth callback does would be nice

1-leo (approved):

packages/ndk/example/bunkers_example.dart
frnandu: You mean in the accounts? It causes circular dependency. Do you gave an idea how to solve that?

1-leo: Hmm i suspected this is the reason. I need a closer look, mabye with a factory

nogringo (approved):

PR #223, opened 2025-08-18
nip46 documentation

nogringo:


frnandu (changes requested):
I would mention nip-46 next to (bunker), and also on the main readme.md add nip-46 (nsecbunker) to the list of Features / what does NDK do.

PR #225, opened 2025-08-20
Add ai guide

nogringo:


nogringo:
> how does it work does claude cli etc read this file (`AI_GUIDE.md`) automatically? if yes we should also add entities so it knows what is returned (or is this too mutch context?)

You should give the file to claude, he do not read it automatically.

1-leo (approved):

PR #232, opened 2025-09-13
Accounts state change listener

nogringo:
Added a stream in the Account class to listen to auth state changes.


1-leo (approved):
👍nice!
Maybe instead of `StreamController` a `BehaviorSubject` so the latest value is always available:
https://pub.dev/packages/rxdart#subjects

1-leo (changes requested):
wait the tests do not check if the actual stream changes, you can do this with 
```
await expectLater(stream, emitsInOrder(<expected>));
```

1-leo (approved):

1-leo:
- flanky test: nip46_event_signer_test

frnandu (approved):

PR #233, opened 2025-09-13
Nip 13 POW

nogringo:
Add the Nip13 class and pow methods for the events.


1-leo (approved):
I thought about doing this with an event signer, but I like this event specific approach.
 We may need to abstract it in the future to allow remote miners or mining in rust. However, this could be done without changing the current API.

1-leo:
Added a test checking validity; not sure why it fails. The test seems ok

frnandu (approved):

PR #238, opened 2025-09-25
Optional pubkey in async signers

nogringo:
Updated the constructor of Nip07EventSigner and Nip46EventSigner to accept a default pubkey.
Making `cachedPublicKey` public.


nogringo:
It's a quick fix waiting the final decision on what to do

PR #253, opened 2025-10-21
feat: rust verifier web

1-leo:
refactor: secp256k1 to rust native dep so it works on web without additional tooling.

I tried to include build scripts so the dev doesn't need to copy the files manually.  Couldn't get it to work.

### DIFF
--- a/packages/rust_verifier/README.md
+++ b/packages/rust_verifier/README.md
@@ -19,7 +19,7 @@ project_root/
 ```
 
 !!!
-The `RustEventVerifier()` object must only be created once! This is a limitation on the web due to the way web workers are handled.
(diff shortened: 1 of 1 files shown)


packages/rust_verifier/README.md
```diff
@@ -3,3 +3,62 @@
...
+```text
+project_root/
+└── web/
+    └── pkg/
+        ├── rust_lib_ndk_bg.wasm
+        └── rust_lib_ndk.js
+```
+
+!!!
+The `RustEventVerifier()` object must only be created once! This is a limitation on the web due to the way web workers are handled.
```
nogringo: Maybe it's unclear, RustEventVerifier must be created once for any platform (even dart)

PR #270, opened 2025-11-03
Unhandled format exception

nogringo:


1-leo (approved):

frnandu (changes requested):
let's not make tests depend on external stuff like echo.websocket.org, otherwise they'll break when external stuff goes away.

1-leo (approved):

frnandu (approved):

PR #271, opened 2025-11-03
Query filters not verified

nogringo:
Step 1: ndk request 1 event of kind 1
Step 2: relay respond with an event of kind 7
Step 3: ndk return the event of kind 7


1-leo:
We discussed this some time ago with no conclustion on how to proceed.
Did you notice specific relay doing this?
We could extend the response cleaner to check again for the query filter.
(the problem does not only apply to kinds)

Issue #227, opened 2025-09-01
ndk.relays appears to operate differently to the bootstrap relays


andrewheadricke:
If I set "bootstrapRelays" (in NdkConfig) to ["wss://nostr.wine"] then I successfully get events when running a query.  However, if I set bootstrapRelays to [] and then run a `ndk.relays.connectRelay(dirtyUrl: "wss://nostr.wine", connectionSource: ConnectionSource.seed);` I get a successful connection established, and the query runs but returns no events.

1-leo:
got it thx. I can reproduce the issue. Will fix it
For now you can  use the `explicitRelays` directly in the request like:
``` dart
final response = ndk.requests.query(
  filters: [
    Filter(limit: 5, kinds: [1]),
  ],
  explicitRelays: ["wss://nostr.wine"],
);
```
That connects does auth and then sends the query
or init with bootstrapRelays as you already did.

PR #278, opened 2025-11-09
fix: ndk.relays appears to operate differently to the bootstrap relays

1-leo:
https://github.com/relaystr/ndk/issues/227


nogringo:
The new tests are passing on my computer so I don't know what the problem is.

PR #332, opened 2025-12-05
refactor: nip01 immutable

1-leo:


frnandu (changes requested):
see previous comment

1-leo:
noticed some weird behavior, if encodingIsolatePoolSize > 1 the tests fail. Looks like our code depends on the order of processed events, suggesting thats its not safe for concurrency. (see tests above)
@frnandu did you have the same problem when using Isolate.spawn() directly?

PR #345, opened 2025-12-22
fix: use isolates in bip340 event verifier

frnandu:


nogringo (approved):

1-leo (commented):
https://github.com/relaystr/ndk/pull/345#issuecomment-3681937081

1-leo (approved):

PR #348, opened 2025-12-22
feat: nip01event immutable light

frnandu:


1-leo (changes requested):
This review is a bit messy. I started by commenting on every occurrence but noticed that it does not make sense. Find and replace will do this in ms 

Its basically

instead of 
``` dart
  final event = Nip01Event(
```

do 

``` dart
final event = Nip01Utils.createEventCalculateId(
```

in our own code so we "future proof " it already.

Has the benefit that we can easily switch to `async` calc id where possible.

PR #350, opened 2025-12-22
feat: rust event verifier using native assets

### LINKED ISSUE #287: Rust event verifier without flutter dependency
### Problem Statement

currently to use the rust event verifier we need flutter_rust_bridge dependency,
would be cool to find a way to compile it without needing flutter

then we could include that in ndk core, and remove the separate package

frnandu:
Using:
- https://pub.dev/packages/native_toolchain_rust
- https://pub.dev/packages/ffigen

We are now able to use rust event verifier implementation using the new native assets without needing flutter nor a separate package 🎉

fixes #287

### DIFF
--- a/packages/rust_verifier/rust_builder/rust/Cargo.lock
+++ b/packages/rust_verifier/rust_builder/rust/Cargo.lock
@@ -61,6 +61,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "4c95c10ba0b00a02636238b814946408b1322d5ac4760326e6fb8ec956d85775"
 
 [[package]]
+name = "arrayvec"
+version = "0.7.6"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "7c02d123df017efcdfbd739ef81735b36c5ba83ec3c59c80a9d7ecc718f92e50"
+
+[[package]]
 name = "atomic"
 version = "0.5.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


frnandu:
since we now have a web verifier, we might be able to allows either use the native assets rust version or the web, not needing anymore the flutter bridge external package, TBC

packages/rust_verifier/pubspec.yaml
```diff
@@ -3,6 +3,9 @@ description: Nostr Development Kit - the most performant lib for all your nostr
 version: 0.5.0-dev.10
 homepage: https://github.com/relaystr/ndk
 
+#discontinued: true
+#replacement: ndk
```
nogringo: Should this be uncommented ?

packages/ndk/lib/src/rust_lib.dart
```diff
@@ -0,0 +1,48 @@
...
+/// Verifies a Schnorr signature.
+///
+/// Returns 1 if valid, 0 if invalid.
+@Native<
+    Int32 Function(
+      Pointer<Utf8>, // pubKeyHex
+      Pointer<Utf8>, // eventIdHex
+      Pointer<Utf8>, // signatureHex
+    )>(symbol: 'verify_schnorr_signature')
+external int verifySchnorrSignatureNative(
```
nogringo: What is this for ? It look unused.

packages/ndk/rust/src/lib.rs
```diff
@@ -0,0 +1,219 @@
...
+            result.push(tag);
+            offset += len as usize;
+        }
+        result
+    } else {
+        Vec::new()
+    };
+
+    // Check id
+    let calc_id = hash_event_data_internal(pub_key, created_at, kind as u16, &tags, content_str);
```
nogringo: In NIP-01 kinds are 32-bit integer. We should use `u32` instead of `u16`.

packages/rust_verifier/pubspec.yaml
frnandu: yep, I commented when I realized that it doesn't work on web and we would probably need it, but now with web verifier it can be discontinued, thanks

packages/ndk/rust/src/lib.rs
frnandu: are you sure? according to nip-01 -> `"kind": <integer between 0 and 65535>,` which is u16 right?

nogringo: you right u16 is perfect

PR #352, opened 2026-01-07
Range based syncing

nogringo:


packages/ndk/lib/domain_layer/entities/filter_coverage.dart
```diff
@@ -0,0 +1,243 @@
...
+      final keys = filter.tags!.keys.toList()..sort();
+      for (final key in keys) {
+        sortedTags[key] = List<String>.from(filter.tags![key]!)..sort();
+      }
+      map['tags'] = sortedTags;
+    }
+
+    final jsonStr = jsonEncode(map);
+    final bytes = utf8.encode(jsonStr);
+    final digest = sha256.convert(bytes);
```
1-leo: this call is expensive, is it possible to make it async and then use isolate from #332

nogringo:
### High-level explanation

This PR introduces a coverage tracking system that records which time ranges have been successfully fetched from each relay for a given filter.

### Queries and streams

The coverage is updated when NDK receive EOSE so it's for queries and streams.
**Should we update the coverage on stream close / ndk destroy ? and / or during the stream ?**

### Optional usecase

Yes it's optional, the dev can use it or not.
**Should we add an option to enable auto coverage on queries / streams ?**

It's the first brick that we need to implement the sync usecase or to do more efficient queries and streams.

### DB structure

Yes, it's independent of events.

### Changing environments (new events in a previous query)

The coverage tracks what we asked for, not what exists.
There is no perfect solution in a decentralized network, the coverage usecase is here to get a picture of what time range we already fetched. And so if we want we can not refecth these ranges.
**Maybe we need to save the fetched date for each time range ?**

frnandu (approved):

PR #357, opened 2026-01-13
Nip42 multi account auth

nogringo:


packages/ndk/lib/domain_layer/usecases/relay_manager.dart
```diff
@@ -37,6 +37,9 @@ class RelayManager<T> {
   /// signer for nip-42 AUTH challenges from relays
   final Accounts? _accounts;
 
+  /// stores the last AUTH challenge per relay for late authentication
+  final Map<String, String> _lastChallengePerRelay = {};
```
frnandu: when does this gets clean up?

frnandu (changes requested):

packages/ndk/lib/domain_layer/entities/ndk_request.dart
```diff
@@ -42,6 +42,10 @@ class NdkRequest {
   /// use cache for write
   bool cacheWrite;
 
+  /// Pubkeys to authenticate with on the relays (NIP-42)
+  /// When set, AUTH events will be sent for each pubkey that has a signer
+  List<String>? authenticateAs;
```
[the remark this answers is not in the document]
frnandu: Signers are quite a low level entity which I think would be best to avoid exposing to the user as much as possible.
Besides, a signer can also not have signing capacity and be read-only.
So I think it's better to expose Account which is a higher level entity exposed to the user.

packages/ndk/lib/domain_layer/usecases/relay_manager.dart
nogringo: It was not clean up.

PR #379, opened 2026-01-22
fix: isolate pool size to use numberOfProcessors

frnandu:
use numberOfProcessors / 4 for isolate pool sizes & add some logging during initialization


1-leo (approved):

1-leo (approved):
- removed the profiler in prod code (its still in /test)

PR #409, opened 2026-02-04
feat: implement RFC 3986 compliant relay URL normalization

### LINKED ISSUE #6: Use Uri for relay url normalization according to RFC 3986

nogringo:
- Add case normalization for scheme and host (Section 6.2.2.1)                
- Add percent-encoding normalization (Section 6.2.2.2)                        
- Add dot segment removal from paths (Section 5.2.4)                          
- Remove default ports: 443 for wss, 80 for ws (Section 6.2.3)                
- Preserve query strings and fragments                                        
- Optimize regex compilation as constants                                     
- Simplify cleanRelayUrls with map/whereType

Closes #6


1-leo:
- move to own file

PR #424, opened 2026-02-06
Fix auth timeout close subscription

nogringo:
AUTH timeout (30s) was longer than query timeout (10s), making it ineffective. Since the WebSocket connection is already established, AUTH OK should arrive within milliseconds. Reduce to 2s for faster failure detection.


nogringo:
There is no bug, the test is wrong. If the user use a broken relay, he will experience timeout and it's expected.

PR #427, opened 2026-02-07
Feat drift cache manager

nogringo:


nogringo:
> What features/advantages does drift db has when compared with other cache impls?

It use web workers and wasm. So Drift will be the most performant web cache.

PR #428, opened 2026-02-07
feat(broadcast): add NIP-09 compliant deletion with e, k, and a tags

nogringo:
- Add event/events params for deletion with k tag
- Add eventAndAllVersions/eventsAndAllVersions params for a tag
- Keep eventId/eventIds for backward compatibility
- Generate a tag for any kind (extension beyond NIP-09)
- Use Set for cache removal deduplication


nogringo:
> this only works in combination with #430 right?

No, but we will need #430 to make this PR clean.

1-leo:
- not perfect solution. Deligate known to delete to some kind of GC

PR #456, opened 2026-02-19
fix(perf): remove calcId from db obj

1-leo:
removes the obsolete calcId from db ojb => faster

grafik

### DIFF
--- a/packages/objectbox/lib/data_layer/db/object_box/db_object_box.dart
+++ b/packages/objectbox/lib/data_layer/db/object_box/db_object_box.dart
@@ -531,13 +531,6 @@ class DbObjectBox implements CacheManager {
     return matchingEventIds ?? <int>{};
   }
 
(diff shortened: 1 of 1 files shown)


packages/objectbox/lib/data_layer/db/object_box/db_object_box.dart
```diff
@@ -485,35 +462,80 @@ class DbObjectBox implements CacheManager {
...
+        return <int>{};
+      }
     }
 
-    // Remove matching events
-    eventBox.removeMany(results.map((e) => e.dbId).toList());
+    return matchingEventIds ?? <int>{};
+  }
+
+  String _normalizeTagKey(String tagKey) {
```
nogringo: This is unused, remove it or use it.

nogringo (approved):

PR #463, opened 2026-02-23
feat: add missing getPublicList

1-leo:
adds:

```dart
  Future<Nip51List?> getPublicList({
    required int kind,
    required String publicKey,
    bool forceRefresh = false,
    Duration timeout = const Duration(seconds: 5),
  }) 
  
  ```
  
  apparently we missed that

### DIFF
--- a/packages/ndk/test/usecases/lists/lists_test.dart
+++ b/packages/ndk/test/usecases/lists/lists_test.dart
@@ -10,41 +10,49 @@ import '../../mocks/mock_relay.dart';
 
 void main() async {
   group('lists', () {
-    KeyPair key0 = Bip340.generatePrivateKey();
-    KeyPair key1 = Bip340.generatePrivateKey();
-    Bip340EventSigner signer0 = Bip340EventSigner(
-      privateKey: key0.privateKey,
-      publicKey: key0.publicKey,
-    );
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by nogringo** [comment]
> what do you suggest? Saving all lists?

```dart
Future<Nip51List?> getPublicList({
  required int kind,
  required String publicKey,
  bool forceRefresh = false,
  Duration timeout = const Duration(seconds: 5),
}) async {
  final signer = Bip340EventSigner(privateKey: null, publicKey: publicKey);

  Nip51List? list =
      !forceRefresh ? await _getCachedNip51List(kind, signer) : null;

  if (list != null) return list;

  final events = await _requests.query(filters: [
    Filter(
      authors: [publicKey],
      kinds: [kind],
      limit: 1,
    )
  ], timeout: timeout).future;

  if (events.isEmpty) return null;

  events.sort((a, b) => a.createdAt.compareTo(b.createdAt));

  await _cacheManager.saveEvent(events.last);
  return await Nip51List.fromEvent(events.last, signer);
}
```

and we can use `await _cacheManager.saveEvents(events);` or `await _cacheManager.saveEvent(events.last);`

nogringo (approved):

frnandu (approved):

PR #465, opened 2026-02-24
fix: preserve tags and custom fields in metadata

### LINKED ISSUE #438: metadata does not store tags/custom content fields
**Describe the bug**
Metadata object does not store tags or custom content:
```
  /// convert to nip01 event
  Nip01Event toEvent() {
    return Nip01Event(
        pubKey: pubKey,
        content: jsonEncode(toJson()),
        kind: kKind,
        tags: [],
        createdAt: updatedAt ?? 0);
  }
```

Is it impossible to use the metadata usecase to fetch profiles that store other content fields?
Say if I want a custom field or tag on the metadata event, and I want to use NDKs caching mechanism for metadata, I suppose that's impossible.

Until recently, https://github.com/nostr-protocol/nips/pull/2216, NIP-39 allowed us to store external identities on "i" tags on metadata, but using NDKs metadata functionality strip that information. 

Are there any plans to store the raw/complete metadata or will it always be the stripped down version?

nogringo:
- Add tags field to Metadata entity
  - Add rawContent field to preserve original JSON
  - Update fromEvent() to store tags and rawContent
  - Update toEvent() to merge with original JSON
  - Add helper methods setCustomField() and getCustomField()
  - Fix constructor to use mutable list for tags

  Fixes #438

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,131 @@
 All notable changes to this project will be documented in this file.
 See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
 
+## 2026-03-04
+
+### Changes
+
+---
+
+Packages with breaking changes:
+
+ - There are no breaking changes in this release.
+
+Packages with other changes:
+
+ - [`ndk_flutter` - `v0.0.2-dev.15`](#ndk_flutter---v002-dev15)
+
+---
+
+#### `ndk_flutter` - `v0.0.2-dev.15`
+
+ - **FIX**: wasm compatible conditional import.
+
+
+## 2026-02-27
+
+### Changes
+
+---
+
+Packages with breaking changes:
+
+ - There are no breaking changes in this release.
+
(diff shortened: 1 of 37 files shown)


packages/ndk/lib/domain_layer/entities/metadata.dart
```diff
@@ -45,20 +57,38 @@ class Metadata {
...
+    String? nip05,
+    String? lud16,
+    String? lud06,
+    this.updatedAt,
+    this.refreshedTimestamp,
+    List<List<String>>? tags,
+    Map<String, dynamic>? content,
+  })  : tags = tags ?? [],
+        content = content ?? {} {
+    // Initialize content with provided known fields
```
frnandu: If both content and individual fields (like name) are passed, the individual fields overwrite content values. This behavior should be documented.

1-leo:
ping @frnandu

Issue #469, opened 2026-02-26
[BUG] Event sources not properly tracked


nogringo:
**Describe the bug**
When the same event is received from multiple relays (during query or broadcast), only the first occurrence is emitted and the `sources` field is not updated to reflect all relays that have the event. This results in incomplete relay hints when encoding events to NIP-19 (nevent) and makes it difficult to know which relays actually have the event.

**Expected behavior**
The `sources` field should contain all relay URLs from which the event was received. When an event is received from multiple relays, the sources should be merged instead of ignoring subsequent occurrences.

**Additional context**
This affects multiple parts of the codebase:

1. **StreamResponseCleaner** (`stream_response_cleaner.dart`): Deduplicates by event ID but ignores subsequent occurrences completely instead of merging sources.
2. **_paginatedQuery** (`requests.dart`): Same issue with `seenEventIds` - duplicates are ignored, sources not merged.
3. **Broadcast operations**: When an event is successfully broadcast to multiple relays, the event object is never updated with the relay URLs that accepted it.
4. 
**Impact:**
- NIP-19 relay hints are incomplete when encoding events
- Pagination may be affected if relying on sources to track queried relays
- Difficult to debug which relays have a given event

PR #470, opened 2026-02-26
fix: event source tracking

nogringo:
Tests to demonstrate https://github.com/relaystr/ndk/issues/469

## Summary by CodeRabbit

* **New Features**
  * Relay content filtering with configurable banned-word blocking.
  * Persistent tracking and merging of event source lists across relays and broadcasts.

* **Improvements**
  * Broadcast flow now updates cached events with per-relay success sources; stream handling supports multiple listeners and avoids duplicate sources.
  * Request/stream processing aggregates sources and re-emits only when source sets grow.

* **Tests**
  * Added tests for broadcast source attribution and multi-relay source merging.

### DIFF
--- a/packages/ndk/lib/domain_layer/usecases/stream_response_cleaner/stream_response_cleaner.dart
+++ b/packages/ndk/lib/domain_layer/usecases/stream_response_cleaner/stream_response_cleaner.dart
@@ -2,6 +2,7 @@ import 'dart:async';
 import '../../../shared/logger/logger.dart';
 import '../../entities/event_filter.dart';
 import '../../entities/nip_01_event.dart';
+import '../../repositories/cache_manager.dart';
 
 /// given a stream  with Nip01 events it tracks the id and adds the one to the provided stream controller \
 /// tracking of the happens in the tracking list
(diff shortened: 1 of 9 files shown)


packages/ndk/lib/domain_layer/usecases/stream_response_cleaner/stream_response_cleaner.dart
```diff
@@ -24,7 +24,7 @@ class StreamResponseCleaner {
     required List<Stream<Nip01Event>> inputStreams,
     required StreamController<Nip01Event> outController,
     required List<EventFilter> eventOutFilters,
-  })  : _trackingSet = trackingSet,
+  })  : _trackingMap = {for (var id in trackingSet) id: {}},
```
1-leo: final instead of var

1-leo (changes requested):

frnandu (changes requested):

frnandu:
because of the  race conditions concerns I'm changing this to a draft, low priority so we can re-discuss possible solutions later

frnandu (changes requested):
race conditions concerns, we need to re-think solution

1-leo:
could use `MutexSimple` for race condition

frnandu:
closed in favor of #679

PR #546, opened 2026-03-28
fix: convert WebSocket message to String on web platforms

pats2sats:
## Problem

On web platforms, `web_socket_client` (which uses `package:web` / `dart:js_interop`) delivers WebSocket text frames as `LegacyJavaScriptObject` rather than Dart `String`s. 

In `RelayManager._handleIncomingMessage`, the `dynamic message` parameter is passed directly to `decodeNostrMsg(String)`, causing:

```
TypeError: Instance of 'LegacyJavaScriptObject': type 'LegacyJavaScriptObject' is not a subtype of type 'String'
```

This crashes the relay connection on Flutter web.

## Fix

Add an explicit `toString()` conversion at the top of `_handleIncomingMessage` before any downstream processing:

```dart
final String msgStr = message is String ? message : message.toString();
```

Then use `msgStr` in place of `message` for all subsequent calls (`decodeNostrMsg`, `runInEncodingIsolate`, `_processDecodedMessage`).

This is a no-op on native platforms (where `message` is already a `String`), and fixes the crash on web.

## Summary by CodeRabbit

* **Bug Fixes**
  * Normalize incoming WebSocket payloads by converting messages to strings before decoding, ensuring consistent size/type handling and more reliable downstream processing.
  * Improve background-task handling so synchronous exceptions are converted into failed futures, making errors reliably surfaced and improving stability of async/background workflows.


PatrickGeyer:
I'm not actually sure that this is the only issue/fix for web NDK stringifying events. I will keep investigating when I see this again, I'll report back.

1-leo:
any updates so far?
I still don't fully understand the issue

PR #571, opened 2026-04-03
feat: add web support to SembastCacheManager

nogringo:
## Summary by CodeRabbit

* **New Features**
  * Cross-platform storage selection: app now picks the appropriate database backend for web vs native.
  * Database path made optional to allow more flexible initialization across platforms.

* **Bug Fixes**
  * Improved validation and clearer errors during database initialization and opening.


1-leo (approved):
ok, as it is. But would prefer via conditional import so the dart:io import is better separated (only on io platforms, not as runtime check)

1-leo (changes requested):
The structure is a bit confusing.
`sembast_cache_manager_web.dart` has a conditional web import

Good would be to have a platform file like `sembast_cache_manager_platform.dart`

with just the exports or the shared code like in file_io:

```
export 'file_io_stub.dart'
    if (dart.library.io) 'file_io_native.dart'
    if (dart.library.js_interop) 'file_io_web.dart';
```

The web import should be wasm compatible
`dart.library.html` => `dart.library.js_interop`

frnandu (approved):

1-leo (approved):

PR #582, opened 2026-04-19
fix: treat 2xx HTTP statuses as success [codex]

pats2sats:
## Summary

- Treat HTTP response statuses in the 2xx range as successful in the shared NDK HTTP request data source.
- Keep Blossom mirror, blob existence checks, and delete result handling aligned with that 2xx success rule.
- Add Blossom tests covering upload 201 Created, mirror 201 Created, and delete 202 Accepted responses.

## Why

Blossom servers may return 201 Created for a successful PUT /upload. The previous strict 200 check caused successful uploads to be reported as failures.

## Validation

- dart analyze lib/data_layer/data_sources/http_request.dart lib/data_layer/repositories/blossom/blossom_impl.dart test/mocks/mock_blossom_server.dart test/usecases/files/blossom_test.dart
- dart test test/usecases/files/blossom_test.dart

### DIFF
--- a/packages/ndk/lib/data_layer/data_sources/http_request.dart
+++ b/packages/ndk/lib/data_layer/data_sources/http_request.dart
@@ -3,6 +3,8 @@ import 'dart:convert';
 import 'package:http/http.dart' as http;
 import 'package:rxdart/rxdart.dart';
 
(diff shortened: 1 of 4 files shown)


1-leo (approved):
can be merged imo

PR #583, opened 2026-04-20
fix: delete ecash proofs when deleting a mint

1-leo:
also deletes the proofs for cashu/wallet delete.

recover possible via restore

## Summary by CodeRabbit

* **New Features**
  * Enhanced mint removal with optional proof deletion control. Users can now choose whether to automatically remove associated proofs when deleting a known mint.


frnandu (approved):

packages/ndk/lib/domain_layer/repositories/wallets_repo.dart
```diff
@@ -48,4 +48,8 @@ abstract class WalletsRepo {
 
   /// Save transactions to storage
   Future<void> saveTransactions(List<WalletTransaction> transactions);
+
+  /// Remove transactions by ID
+  /// should only be used for deleting a wallet
+  Future<void> removeTransactions(List<String>? transactionIds);
```
nogringo, line 55: Should we replace `List<String>?` by `List<String>` ?
Or does when transactionIds is null, does it means: "Delete all transactions" ?

nogringo (changes requested):
`cashu_wallet_provider.dart:67` should be awaited

packages/ndk/lib/domain_layer/repositories/wallets_repo.dart
1-leo: good point. 
null means delete all. Will update it in the comment

nogringo (approved):

PR #588, opened 2026-04-23
[codex] add configurable signer and NIP-44 cryptography

pats2sats:
## What changed

This PR adds two small configuration seams for NDK cryptography:

- `eventSignerFactory` in `NdkConfig`, used whenever NDK needs to construct a signer internally
- `nip44Cryptography` in `NdkConfig`, used by the default local `Bip340EventSigner` for NIP-44 encrypt/decrypt

Together these changes let downstream apps swap in faster or platform-backed cryptography without carrying local patches across bunker, gift-wrap, NWC, Blossom, and other internally-created signer paths.

## Why

NDK previously hardcoded `Bip340EventSigner` construction in several internal flows, and that signer hardcoded the built-in NIP-44 implementation. That made it difficult for downstream apps to reuse an existing optimized signer or NIP-44 backend consistently.

The main impact is that apps can now:

- keep local private-key flows on a custom signer implementation
- keep bunker session-local crypto on the same fast path
- override local NIP-44 encrypt/decrypt independently when they do want to use the default signer type

## Impact

- No behavior change for existing users unless they opt into the new config fields
- Existing defaults continue to use the built-in signer and NIP-44 implementation
- Downstream apps get a single config-level way to inject custom cryptography

## Validation

- `dart test test/usecases/accounts_test.dart test/signers/nip46_event_signer_test.dart test/usecases/gift_wrap/gift_wrap_test.dart`
- `dart test test/signers/bip340_event_signer_test.dart test/usecases/accounts_test.dart`
- `dart analyze` on the touched config/signer files

### DIFF
--- a/packages/ndk/lib/data_layer/repositories/cryptography/default_nip44_cryptography.dart
+++ b/packages/ndk/lib/data_layer/repositories/cryptography/default_nip44_cryptography.dart
@@ -1,0 +1,24 @@
+import '../../../domain_layer/repositories/nip44_cryptography.dart';
+import '../../../shared/nips/nip44/nip44.dart';
+
+class DefaultNip44Cryptography implements Nip44Cryptography {
+  const DefaultNip44Cryptography();
+
+  @override
+  Future<String> encrypt({
+    required String plaintext,
+    required String privateKey,
+    required String publicKey,
+  }) {
+    return Nip44.encryptMessage(plaintext, privateKey, publicKey);
+  }
+
+  @override
+  Future<String> decrypt({
+    required String ciphertext,
+    required String privateKey,
+    required String publicKey,
+  }) {
+    return Nip44.decryptMessage(ciphertext, privateKey, publicKey);
+  }
+}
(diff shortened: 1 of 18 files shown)


PatrickGeyer:
Hey folks, raw NDK was unusable on web due to slow event validation and signing. I really needed a way to pass in custom cryptography, especially when using things like bunker or giftwraps which need to construct and deconstruct a bunch of events. Don't expect this to be merged, just think it's a reasonable feature request!

1-leo:
Hey thx for this feature request. We already identified the issue and there is `WebEventVerifier` available in `ndk_flutter`.
`WebEventVerifier` uses native web crypto apis and is way faster then any wasm bridge.
We will update the documentation and let you know.

To make it more convenient, we are building a hybrid signer that uses the best default per platform.

cc: @nogringo

1-leo:
https://github.com/relaystr/ndk/pull/604 https://github.com/relaystr/ndk/pull/602 https://github.com/relaystr/ndk/pull/601

PR #598, opened 2026-05-02
[codex] Preserve NIP-46 signed created_at

pats2sats:
## What changed

- Copy `created_at` from the NIP-46 `sign_event` response in addition to `id` and `sig`.
- Keep the original local event body (`kind`, `content`, `tags`, and `pubkey`) as the returned event body.
- Add regression coverage for a remote signer that returns a later `created_at`, plus a test proving remote content changes are not accepted as the returned body.

## Why

NIP-46 `sign_event` returns a complete signed event. Some bunkers can normalize or overwrite `created_at` while signing, especially when a request is delayed. The returned `id` and signature are then valid for the event using that returned timestamp.

Previously NDK copied only `id` and `sig` onto the original local event. If the bunker returned a newer `created_at`, the resulting event body no longer matched the signed hash and relays rejected it as an invalid id/signature.

This patch copies only the timestamp needed to preserve the signed hash while keeping the event content/tags/kind requested by the app as the authoritative returned body.

## Validation

- `dart test test/signers/nip46_event_signer_test.dart`
- `dart analyze` exits 0, with existing info-level lints in examples/tests.

### DIFF
--- a/packages/ndk/lib/data_layer/repositories/signers/nip46_event_signer.dart
+++ b/packages/ndk/lib/data_layer/repositories/signers/nip46_event_signer.dart
@@ -257,9 +257,14 @@ class Nip46EventSigner implements EventSigner {
       request: request,
       event: event,
     );
-    final signedEvent = jsonDecode(signedEventJson);
 
-    return event.copyWith(id: signedEvent["id"], sig: signedEvent["sig"]);
+    final signedEvent = Nip01EventModel.fromJson(jsonDecode(signedEventJson));
+
+    return event.copyWith(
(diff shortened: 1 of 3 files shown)


frnandu:
If you submit PRs and want us to review/merge it, mark it as ready for review, otherwise while in `Draft` we'll not know if it's ready to be reviewed or not @PatrickGeyer

nogringo:
Nice find ! I improved your in https://github.com/relaystr/ndk/pull/600 it now keep the signed event as is so it let full freedom to the bunker.

PR #601, opened 2026-05-06
feat: add PlatformEventVerifier

nogringo:
- Add PlatformEventVerifier that auto-selects the best verifier per platform (WebEventVerifier on web, RustEventVerifier on native)
- Update docs to recommend PlatformEventVerifier for Flutter apps

### DIFF
--- a/doc/concepts/performance.md
+++ b/doc/concepts/performance.md
@@ -9,4 +9,6 @@ Even splitting the users filters into smaller relay tailored filters if we know 
 **compute**\
 Right now the most compute intensive operation is verifying signatures. \
 We use the cache to determine if we have already seen the event and only if it is unknown signature verification is done. \
-To make the operation as optimized as possible we strongly recommend using `RustEventVerifier()` because it uses a separate thread for verification.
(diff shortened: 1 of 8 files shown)


frnandu (approved):

1-leo (changes requested):
can we do this by checking the api vs the platform, the typical conditional import with a stub like in file_io (in this case a factory):

export 'file_io.dart';

/// Exports the correct FileIO implementation based on platform.
export 'file_io_stub.dart'
    if (dart.library.io) 'file_io_native.dart'
    if (dart.library.js_interop) 'file_io_web.dart';

/// Exposes `createFileIO()` implemented per platform.
export 'file_io_factory_stub.dart'
    if (dart.library.io) 'file_io_factory_native.dart'
    if (dart.library.js_interop) 'file_io_factory_web.dart';

1-leo (approved):

1-leo:
- [ ] rename platform event verifier to ndk_event_verifier

frnandu (approved):
after renaming to NdkEventVerifier

PR #602, opened 2026-05-07
feat: web/ndk event signer

nogringo:
## User Story

As a developer I want to use the fastest signer available so I don't have to write platform specific code myself, saving time

## Release Notes

* **New Features**
  * WebEventSigner: Fast web-based event signing and encryption/decryption using Web Crypto APIs for BIP-340 signing and NIP-04/NIP-44 cryptography
  * NdkEventSigner: Automatic platform selection between WebEventSigner (web) and Bip340EventSigner (native)
  * NdkEventVerifier: Replaces PlatformEventVerifier with improved automatic platform selection for verification

* **Documentation**
  * Updated guides and concepts to document new signer/verifier types and platform selection behavior


packages/ndk_flutter/lib/signers/platform_event_signer.dart
```diff
@@ -0,0 +1,9 @@
+/// Platform-aware event signer that uses the best implementation
+/// for the current platform.
+///
+/// On web: uses [WebEventSigner] (fast JS crypto via @noble/curves)
+/// On native: uses [Bip340EventSigner] (pure Dart)
+library;
+
+export 'src/platform_event_signer_stub.dart'
+    if (dart.library.js_interop) 'src/platform_event_signer_web.dart';
```
1-leo: the IO import is missing here, or does this happen elsewhere?

1-leo (changes requested):

packages/ndk_flutter/lib/verifiers/assets/nostr_crypto.src.js
```diff
@@ -0,0 +1,336 @@
...
+  return okm.slice(0, length);
+}
+
+function equalBytes(a, b) {
+  if (a.length !== b.length) return false;
+  for (let i = 0; i < a.length; i++) {
+    if (a[i] !== b[i]) return false;
+  }
+  return true;
+}
```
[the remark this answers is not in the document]
1-leo: @nogringo 👆

packages/ndk_flutter/lib/signers/platform_event_signer.dart
nogringo: @1-leo No dart:io import is needed here.

`platform_event_signer.dart` uses Dart's conditional exports to pick the implementation at compile-time.
No runtime platform check required.

Do you still want the 2nd `if dart io` ?

1-leo: then i dont understand it, how is the `Bip340EventSigner` used?
I would assume it works like this:
```
export 'src/platform_event_signer_stub.dart'
    if (dart.library.js_interop) 'src/platform_event_signer_web.dart';
    if (dart.library.io) '<path_to_Bip340EventSigner>':

```

Another thought: How does this work with `Nip46EventSigner` I assume when remote is active, the platform signer is skipped, right?

nogringo: We can have multiple signers used at the same time, Nip46 for one account and Nip07 for an other.

Also, Nip46 has a local Bip340EventSigner that should be WebEventSigner on web. So the platform signer should not be skipped.

1-leo:
- [x] rename platform event signer to ndk_event_signer

frnandu (approved):
after renaming to NdkEventSigner

1-leo (changes requested):
rename, integrate with https://github.com/relaystr/ndk/pull/604

1-leo (approved):
looks ok. Just one confusing naming thing:
there is `ndk_event_signer_web.dart` and `web_event_signer_web.dart`
The naming of the files is a bit confusing internally one is the factory and one the `WebEventSigner` for me it would be ok if both are in one file `event_signer_web.dart`

PR #606, opened 2026-05-10
feat: blob cache manager

nogringo:
Like Events Cache Manager but for blobs

## Summary by CodeRabbit

# Release Notes

* **New Features**
  * Added local blob caching for Blossom file operations, reducing network calls for repeated blob access.
  * Introduced configurable cache managers with persistent and opt-out options.
  * Added `cacheWrite` parameter to blob operations for granular caching control.

* **Documentation**
  * Enhanced persistence guide with blob cache configuration examples.
  * Updated Blossom usage documentation with local cache behavior details.

* **Tests**
  * Added comprehensive test coverage for blob caching functionality.

[Review Change Stack](https://app.coderabbit.ai/change-stack/relaystr/ndk/pull/606)

### DIFF
--- a/doc/guides/persistence.md
+++ b/doc/guides/persistence.md
@@ -3,7 +3,18 @@ icon: database
 order: 98
 ---
 
-Ndk comes with several database offerings. The simplest is the `MemCacheManager` which is an in-memory cache. This is useful for testing and small applications.
+NDK keeps two local stores side by side:
+
+- the **event cache** (`CacheManager`) — Nostr events, contact lists, metadata, NIP-05 records, relay sets, fetched ranges, etc.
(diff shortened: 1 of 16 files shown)


frnandu (changes requested):
Confirm that there are some kind of memory limits in the idb_shim lib , otherwise the cache will grow without bounds. Provide documentation on how to configure such limits

1-leo (changes requested):
code looks good.

I would like to remove the idb_shim dependency.
Maybe ship ndk just with the `NoopBlobCacheManager` and use a seperate package (same pattern as the other db implementations), or maybe its possible to talk to the storage api (platform) with the tmp folder directly (preferred if possible)

- [ ] there needs to be a cleanup system so it does not grow indefenently

nogringo:
https://pub.dev/packages/blossom_cache

PR #644, opened 2026-06-03
Feat/nip37 private storage relays

nogringo:
This pr only add a high level api to get the private relays list

## Release Notes

* **New Features**
  * Added ability for logged-in users to fetch their private storage relay list.

* **Documentation**
  * Updated documentation to explain private storage relay functionality and usage.

* **Tests**
  * Added test coverage for private storage relay retrieval.

### DIFF
--- a/packages/bc_ur/lib/ur.dart
+++ b/packages/bc_ur/lib/ur.dart
@@ -19,9 +19,7 @@ class UR {
   @override
   bool operator ==(Object other) {
     if (identical(this, other)) return true;
-    return other is UR && 
-           other.type == type && 
-           _listEquals(other.cbor, cbor);
+    return other is UR && other.type == type && _listEquals(other.cbor, cbor);
   }
 
   @override
--- a/packages/ndk/lib/domain_layer/entities/nip_37.dart
+++ b/packages/ndk/lib/domain_layer/entities/nip_37.dart
@@ -1,0 +1,4 @@
+class Nip37 {
+  /// Relay list for private content.
+  static const int kPrivateStorageRelays = 10013;
+}
--- a/packages/ndk_flutter/lib/data_layer/repositories/signers/amber_event_signer.dart
+++ b/packages/ndk_flutter/lib/data_layer/repositories/signers/amber_event_signer.dart
@@ -13,9 +13,7 @@ class _PendingRequestEntry {
 }
 
 /// amber (external app) https://github.com/greenart7c3/Amber singer
-class AmberEventSigner
-    with ConcurrencyLimiterMixin
-    implements EventSigner {
+class AmberEventSigner with ConcurrencyLimiterMixin implements EventSigner {
   final AmberFlutterDS amberFlutterDS;
 
   final String publicKey;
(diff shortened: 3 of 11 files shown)


frnandu:
```
jsonDecode on null decrypt will throw (user_relay_lists.dart, _extractPrivateStorageRelayUrls)

    dart
    final json = await signer.decryptNip44(
      ciphertext: event.content,
      senderPubKey: event.pubKey,
    );
    final tags = jsonDecode(json ?? '') as List<dynamic>;


    If decryptNip44 returns null (wrong key, corrupted content, etc.), json ?? '' becomes '', and jsonDecode('') throws FormatException. The t
ry/catch in the caller swallows it and returns null, which works -- but it conflates "decryption failed" with "malformed payload". A user seei
ng null has no idea which. Consider checking json == null explicitly and returning null or throwing a meaningful error before attempting jsonD
ecode.
```

packages/bc_ur/lib/ur.dart
```diff
@@ -19,9 +19,7 @@ class UR {
   @override
   bool operator ==(Object other) {
     if (identical(this, other)) return true;
-    return other is UR && 
-           other.type == type && 
-           _listEquals(other.cbor, cbor);
+    return other is UR && other.type == type && _listEquals(other.cbor, cbor);
```
1-leo, line 22: is this the default dart formatter?

packages/ndk_flutter/lib/data_layer/repositories/signers/amber_event_signer.dart
1-leo, line 1: this signer is deprecated => merge from master

1-leo (approved):
There are some formatting changes that cause merge conflicts. 
Are you using the default dart formatter or did a dart version change it?

packages/ndk/lib/domain_layer/entities/nip_37.dart
```diff
@@ -0,0 +1,4 @@
+class Nip37 {
+  /// Relay list for private content.
+  static const int kPrivateStorageRelays = 10013;
```
frnandu, line 3: This kind is not defined in nip-37, but in nip-51.
The fact that draft relays are used in nip-37 doesn't make it as the place that this kind is defined.

frnandu (changes requested):
I don't think we need to have this PR at all.
Using `ndk.lists.getPublicList` it should be possible to fetch list for kind `10013`.
Just add this `10013` kind to `Nip51List` as `kDraftRelays` which is missing there, and you should be good to use it.
wdyt @1-leo ?

1-leo:
```
ndk.<somename>.getPrivateUserRelays()

ndk.lists.getPrivateUserRelays()

ndk.lists.relays.getDraftUserRelays()

#new usecase not the current userRelayLists
# internally uses lists usecase
ndk.userRelayLists.setPriateRelays()

ndk.drafts.getDraftsRelays()
ndk.drafts.setDraftRelays()



ndk.lists.getSetByName(<myname>, <kindForList>)
# => user need to read spec

```

plan for now:
new class (name tbd) with helper methods for relay related list stuff `getDraftsRelays()`, `getPrivateRelays()`
in the next call, decide for name and `ndk.<someUsecase>.<method>` or `ndk.lists.<someName>.<method>`

1-leo:
https://github.com/nostr-protocol/nips/pull/2404

=> `ndk.lists.getPrivateUserRelays()`

PR #648, opened 2026-06-06
Fix/mock relay live subscriptions

nogringo:
## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed MockRelay's EVENT handling to correctly suppress broadcasting for replaceable and addressable events when they should not replace stored data.

* **Tests**
  * Added integration test for MockRelay live subscriptions, verifying cross-instance event broadcasting and subscription delivery.


frnandu (approved):
1. Test coverage gap: The test only covers the happy path (text note broadcast). The core behavioral change -- suppressing broadcast for stale replaceable/addressable events -- is not tested. Consider adding a test that:
       - Publishes a replaceable event (e.g. kind 10000)
       - Publishes an older version of the same event
       - Asserts the subscriber only receives the first event, not the rejected stale one

1-leo (approved):

PR #657, opened 2026-06-10
Feat/nip55 external signer

nogringo:
## Release Notes

* **New Features**
  * Added full NIP-55 external signer support for authentication and signing/encryption/decryption via compatible signer apps.
  * Integrated signer-app availability checks and targeted signer launching.

* **UI Updates**
  * Replaced “Login with Amber” with “Login with signer app” across all supported languages.
  * Added an optional signer-app login entry to the login widget.

* **Bug Fixes**
  * Improved startup/account restoration resilience for external-signer flows.
  * Fixed QR rendering and QR quiet-zone configuration in wallet dialogs.

* **Infrastructure**
  * Updated Android/Kotlin and Flutter plugin configuration; removed Amber Flutter dependency.


nogringo:
Metioned in nip55 and not implemented:

- [ ] decrypt_zap_event (not in ndk abstract signer)
- [ ] batch requests
- [ ] gzip for large requests
- [ ] web signing via callbackUrl

1-leo (approved):
tested and looks good!

Because its a braking change we should publish a migration guide in our docs (basically just renaming stuff) wdyt?

Also `amberflutter` is still mentioned in some codedocs and skills.md

frnandu (changes requested):
- Sample app AmberPage class name -- The class is still called AmberPage and the variable is amber. Since this is now generic, renaming to SignerAppPage / signerApp would be cleaner

- isAppInstalled hardcodes Amber's package name
       - nip55_signer.dart:77 always checks com.greenart7c3.nostrsigner. The whole point of this PR is to be signer-generic, yet install detec
tion is Amber-specific. If the user has Primal/Aegis but not Amber, isAppInstalled() returns false, and the UI redirects to Amber's GitHub ins
tead of letting them proceed.
       - The Kotlin side has isExternalSignerInstalled() which checks the nostrsigner: scheme (generic), but the Dart side doesn't call it wit
h packageName: null -- it always passes the Amber package.

- isAppInstalled logic in Kotlin has surprising OR
       - DartNdkPlugin.kt:186: isPackageInstalled(...) || isExternalSignerInstalled(_context) -- if the specific package is not installed but ANY external signer is, this returns true. For isAppInstalled called with a specific package, this seems misleading. It may cause the UI to think a specific signer is available when it's actually a different one.

frnandu (approved):

PR #662, opened 2026-06-15
doc: adr broadcast offline

1-leo:
ADR for offline broadcast refactor

## Summary by CodeRabbit

* **Documentation**
  * Added an Architecture Decision Record documenting the Offline Broadcast approach, including system design, broadcast and eviction policies, retry and error-handling requirements, performance considerations, and accepted alternatives.


frnandu:
This will pretty quickly be out of date, and I don't want to still need to update such docs along the way, so I would propose to not keep such doc that will be anyhow out-of-date once everything is in place, unless you want to come back to this and update it to the final implementation details @1-leo

PR #677, opened 2026-06-19
Fix/mock relay tag filtering

nogringo:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved tag-based event filtering in relay query responses, so results now respect multiple tag constraints more consistently.
  * Fixed handling of tag filters so malformed tag entries are safely ignored instead of causing failures.

### DIFF
--- a/packages/ndk/test/mocks/mock_relay.dart
+++ b/packages/ndk/test/mocks/mock_relay.dart
@@ -449,6 +449,10 @@ class MockRelay {
         }
       }
 
+      eventsForThisFilter = eventsForThisFilter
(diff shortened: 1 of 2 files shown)


frnandu (approved):
Suggestions (non-blocking)

1. Add regression tests for #p/#e — these were the previously-supported tags; verify they still work through the generic path.
2. Test multi-tag AND — {\"#r\": [\"a\"], \"#p\": [\"<pubkey>\"]} to confirm both must match.
3. Test malformed tags — single-element tag like ['r'] should be safely skipped (the old crash scenario).

1-leo (approved):

PR #689, opened 2026-06-30
Fix/cache served query completes immediately

nogringo:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved request handling so cached queries finish cleanly once all results have been served, avoiding unnecessary waiting on the network.
  * Queries filtered by event ID now return cached results immediately and complete faster.

* **Tests**
  * Added coverage for ID-based cache queries to verify correct results and fast completion from local storage.


frnandu (approved):
nice catch

1-leo:
question: remove filter for replaceable events

1-leo (approved):

PR #679, opened 2026-06-23
feat: local first

frnandu:
This work established the core local-first delivery and cache model for NDK, focusing first on correctness and high-level behavior before broader sync work.

### Key changes

- Added persistent per-event delivery state in cache, so unpublished events and relay delivery progress survive restarts.
- Split event provenance from `Nip01Event.sources`; source relays are now loaded on demand through `CacheManager` APIs instead of always living on the event object.
- Deprecated `Nip01Event.sources`.
- Added separate storage for relay delivery targets to avoid unsafe concurrent mutation of shared target lists.
- Added persisted interactive-signing state in `EventDeliveryRecord`, so local-first work can survive while an event is still unsigned and waiting on an external signer flow.
- Moved `Metadata` and `ContactList` fully onto the generic authoritative event store, keeping them as convenience accessors/views instead of separate primary cache silos.
- Deprecated dedicated `CacheManager` write/remove APIs for `Metadata` and `ContactList` in favor of generic event persistence.
- Introduced delivery policy classification for retries:
  - replaceable events use latest-state-only behavior
  - ephemeral events do not retry
  - deletion/control events use faster retry backoff
  - permanent vs transient relay failures are classified from NIP-style prefixes and common relay messages
- Added periodic due-retry handling for still-connected relays, so transient `OK false` failures such as rate limits are retried even when no relay connectivity change happens.
- Extended `EventSigner` with local-first scheduling hints for delayed signing:
  - `requiresInteractiveSigning`
  - `requiresSignerNetwork`
  - `signerTransportRelayUrls`
- Extended `PendingBroadcastDelivery` so it can retry signing as well as relay delivery:
  - unsigned interactive-signer events stay queued locally
  - network-backed signers are retried when their transport relays reconnect
  - successful delayed signing immediately triggers publish flush to connected target relays
  - network-backed signers use slower retry backoff than local interactive signers
- Added event kind classification helpers in `EventKindClassification` for replaceable, addressable, regular replaceable, parameterized replaceable, and ephemeral kinds.
- Replaced blunt numeric event-kind checks with existing constants plus shared helpers, including `Metadata.kKind`, `ContactList.kKind`, `10002`, and `ChannelMetadata.kKind` for kind `41`.
- Improved cache visibility rules so reads consistently enforce:
  - replaceable winner selection
  - expiration filtering
  - NIP-09 deletion/tombstone suppression
[...]

PR #692, opened 2026-07-06
fix(ndk_flutter): migrate Android plugin to Built-in Kotlin

nogringo:
Apply the Kotlin Gradle Plugin only on AGP < 9 so Flutter 3.44+ stops flagging ndk_flutter as applying KGP and future Flutter versions keep building. The conditional stays on one line so Flutter's source-level detection does not re-trigger the warning. Replace the deprecated kotlinOptions block with kotlin { compilerOptions }, which also works under AGP 9. Remains compatible with Flutter < 3.44.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Android build compatibility with newer tooling versions.
  * Updated Kotlin compilation settings to use a consistent JVM 17 target.

### DIFF
--- a/packages/ndk_flutter/android/build.gradle
+++ b/packages/ndk_flutter/android/build.gradle
@@ -21,7 +21,11 @@ allprojects {
 }
 
 apply plugin: 'com.android.library'
-apply plugin: 'kotlin-android'
+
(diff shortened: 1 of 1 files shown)


frnandu:
I did this on #679 , please review it so we can merge it and move on to work on other things.

PR #698, opened 2026-07-16
test: add metadata NIP-65 relay discovery regression

nogringo:
## Summary by CodeRabbit

- **Improvements**
  - Metadata loading now follows an author’s preferred write relay when available, improving profile information retrieval.

- **Bug Fixes**
  - Relay responses now correctly return multiple event types from combined queries, including metadata, contacts, relay lists, and replaceable events.

- **Tests**
  - Added coverage for metadata retrieval through author write relays across supported relay engines.
  - Added validation for combined event-kind queries.

### DIFF
--- a/packages/ndk/lib/domain_layer/usecases/metadatas/metadatas.dart
+++ b/packages/ndk/lib/domain_layer/usecases/metadatas/metadatas.dart
@@ -12,6 +12,7 @@ import '../../repositories/event_signer.dart';
 import '../accounts/accounts.dart';
 import '../broadcast/broadcast.dart';
 import '../requests/requests.dart';
+import '../user_relay_lists/user_relay_lists.dart';
 
 /// nostr metadata usecase
 class Metadatas {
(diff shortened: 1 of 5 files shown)


1-leo (commented):
This works, however in the initial concept of NDK gossip the engines are responsible for calling the right relays. 
This shortcuts this by passing explicit relays.
Not sure if we want this, because for consistency we would need it in other use cases as well, making the discovery part in the engines obsolete.

nogringo:
As in the tests actually both engines were failing

1-leo:
@1-leo check test use await for the first gossip discovery

frnandu (changes requested):
packages/ndk/lib/domain_layer/usecases/metadatas/metadatas.dart:L59: 🔴 bug: relay-list discovery ignores loadMetadata's idleTimeout; a 100 ms request took 4.2 s because discovery waited on relay connection first. Bound discovery + metadata lookup by caller deadline or propagate remaining timeout.

PR #702, opened 2026-08-07
docs: add ADR for local-first reads

nogringo:
Defines NdkDataResponse<T> and NdkValue<T> as the return format for reads that render from cache first and refine when relays answer, with the private relay list (NIP-37 kind 10013) as first use.

## Summary by CodeRabbit

* **Documentation**
  * Added an architecture decision record describing local-first read responses.
  * Documented how cached results are delivered before relay-confirmed updates.
  * Defined result metadata distinguishing cached values, relay values, loading states, and confirmed empty results.
  * Documented future-based final results and stream-based error reporting.
  * Included private relay-list reads as the initial application, including the distinction between an empty list and no list.

connected to:
https://github.com/relaystr/ndk/pull/721/changes

### DIFF
--- a/doc/library-development/ADRs/local-first-reads.md
+++ b/doc/library-development/ADRs/local-first-reads.md
@@ -6,7 +6,7 @@ Title: Local-first reads - return format for reads that render without waiting o
 
 proposed
 
-Updated on 05-08-2026
+Updated on 2026-08-17
 
 ## contributors
 
@@ -26,38 +26,38 @@ A read must also never let one state degrade into another. "not known yet" and "
 must not look like "does not exist", because that is what an app acts on to create the missing
 data, and for a replaceable event that overwrites what was already there.
 
+## Scope
+
+This is the return format of typed high-level reads: one read, one value, refined until it is
+relay-confirmed. It does not redefine `requestNostrEvent` and `NdkResponse`, and it is not a
+lifecycle for long-lived subscriptions.
+
+Per-relay provenance stays out of it: which relays hold an event is a cache concern, already
+answered by `CacheManager.addEventSource(s)` and `loadEventSources()`. Read-specific metadata
+belongs in `T`, so a read that needs more than the origin declares a `T` that carries it.
+
 ## Main Proposal
 
 ```dart
 class NdkDataResponse<T> {
   /// Emits a `cache` value first, then every newer `relays` value as it
-  /// arrives, even when it holds the same value as the cache (TBD).
-  /// Closes after EOSE or timeout.
+  /// arrives. Closes after EOSE or timeout.
   final Stream<NdkValue<T>> stream;
 
-  /// The relay-confirmed value, so the last emitted one.
-  Future<T> get future;
+  /// The last emitted value, relay-confirmed unless the read concludes on cache.
+  Future<NdkValue<T>> get future;
 }
 
 class NdkValue<T> {
-  final T value;
+  final T? value;
   final DataOrigin origin;
 }
 
 enum DataOrigin { cache, relays }
 ```
 
-NOTE: Does not cover ID queries where cache response is sufficient TBD
-
-A `cache` value is always emitted first, even when nothing is cached.
-
-`stream` is backed by a `BehaviorSubject`, so `stream` and `future` can both be consumed and a
-listener attached late still receives the latest value.
-
-`NdkValue` may later carry `createdAt`, `receivedAt` and `hasPendingWrites`. Only metadata that
-makes sense for every read belongs there, anything specific to one read belongs in `T`.
-
-When the value can be absent, `T` is nullable and the origin disambiguates `null`:
+A `cache` value is always emitted first, even when nothing is cached, so `value` is nullable
+for every read and the origin disambiguates `null`:
 
 | emission | meaning |
 | --- | --- |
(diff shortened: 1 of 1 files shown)


1-leo:
@nogringo why is a generic used for value
```
class NdkValue<T> {
  final T value;
  final DataOrigin origin;
}
```
Its always `Nip01Event` or do you want to also use it for use cases with special return types like lists and metadata?

Added TBD for stuff where I am unsure of

nogringo:
Yes, I used `NdkValue<T>` for others use cases with special return type

frnandu:
Its main weaknesses for your use case:

  1. It cannot identify individual relays

  The proposed metadata is only:

  enum DataOrigin { cache, relays }

  That answers “cache or some relay,” but not:

  - which relay returned the event;
  - whether several relays returned it;
  - whether one relay failed while another succeeded;
  - which relay supplied which event.

  It needs at least relayUrl, or preferably Set<String> relayUrls in an aggregated result.

  2. It does not cover query() or subscription()

  The ADR’s first use is getPrivateUserRelays(), a higher-level single-value read. It explicitly leaves ID queries unresolved and does not
  describe arbitrary multi-event queries or ongoing subscriptions. The review calls out the unresolved ID-query contract directly.
  (https://github.com/relaystr/ndk/pull/702#discussion_r3740641765)

  Applying it to NdkResponse requires changing the internal cache/network merge. Right now both inputs become plain Nip01Events and duplicate
  IDs are discarded in packages/ndk/lib/domain_layer/usecases/stream_response_cleaner/stream_response_cleaner.dart:38.

  3. One enum value cannot retain combined provenance

  An event may be:

  - initially served from cache;
  - subsequently confirmed by relay A;
  - subsequently confirmed by relay B.

  A sequence of NdkValues can express that temporarily, but a late BehaviorSubject subscriber sees only the latest value, while Future<T>
  strips the origin entirely. The resulting value cannot answer “cache and network, and which relays?”

  4. Equal cache/relay values remain unresolved

  The ADR says equal relay values may still emit, but marks that decision TBD. For provenance, they must emit: an identical event from a relay
  is still new information because it confirms availability on that relay.

  The current request cleaner deduplicates by event ID, so wrapping its output afterward would be too late—the relay observation has already
  been discarded.

  5. Relay winner selection is underspecified

  The ADR defines future as the “last emitted” relay value. For replaceable events, the last relay response may contain an older event. The
  final value must use NIP-01 replacement ordering, not arrival order. This was identified in review as a major correctness issue.
  (https://github.com/relaystr/ndk/pull/702#discussion_r3736495977)

  6. Its loading/null state is not type-safe

  It promises an initial cache emission even on a miss, but NdkValue<T>.value cannot be null when T is non-nullable. The model needs either
  nullable values everywhere or, preferably, explicit sealed states such as loading, value, confirmedEmpty, and failure. The review flags this
  mismatch. (https://github.com/relaystr/ndk/pull/702#discussion_r3736495989)

  7. Failure and completion semantics are incomplete

  The ADR does not fully specify:

  - how future completes when no relay is reachable;
  - whether cache decryption errors prevent relay refinement;
  - how partial relay failures are handled;
  - whether timeout returns the best known result or throws.

  These gaps are also documented in review. (https://github.com/relaystr/ndk/pull/702#discussion_r3736495986)

  8. Query and subscription semantics differ

  “Closes after EOSE or timeout” fits a query, but not a long-lived subscription. A subscription needs observation-level events and should
  remain open after initial relay EOSE. The ADR currently treats local-first reads as a single general lifecycle.

  Recommended direction

  Keep PR #702’s local-first concept for higher-level APIs, but introduce a lower-level observation type for requests:

  enum NdkEventOrigin { cache, relay }

  class NdkEventObservation {
    final Nip01Event event;
    final NdkEventOrigin origin;
    final String? relayUrl;
    final DateTime receivedAt;
  }

  Then expose both:

  class NdkResponse {
    // Every cache/relay observation, including duplicate event IDs.
    final Stream<NdkEventObservation> observations;

    // Existing backward-compatible, event-ID-deduplicated projection.
    final Stream<Nip01Event> stream;
  }

  For completed queries, aggregate observations into something like:

  class NdkEventResult {
    final Nip01Event event;
    final bool servedFromCache;
    final Set<String> relays;
  }

  That preserves backward compatibility while answering exactly:

  - Was this event served from cache?
  - Was it observed from the network during this request?
  - Which relays returned it?
  - Was it both cached and relay-confirmed?

  So I would treat PR #702 as a useful semantic starting point, but not merge its proposed response shape as the solution for raw queries/
  subscriptions without first adding per-relay observations and resolving the outstanding lifecycle contracts.

nogringo:
I updated the ADR, but part of this measures it against a use case it does not claim to cover.

Updated: a scope section (typed high-level reads, not `requestNostrEvent` / `NdkResponse`, not a subscription lifecycle), "newer" defined as NIP-01 replacement ordering, `NdkValue.value` nullable with `future` returning the whole `NdkValue`, so the loading and cache-miss states are representable for every `T` (6), and both TBDs resolved: an equal relay value emits once because the origin change is the confirmation, and an id hit on a regular event is terminal on `cache`.

5 and 7 were already answered by the contract: the stream emits only *newer* values, so the last emitted one is the NIP-01 winner by construction, and `future` is that last emission, hence the same error channel. Nothing separate to specify.

1, 3 and 8 are the extension point the ADR already states: `NdkValue` carries only what makes sense for every read, read-specific metadata goes in `T`. Per-relay provenance is exactly that, and NDK already has it in the cache: `CacheManager.addEventSource(s)` / `loadEventSources()`.

1-leo:
I guess the main question is do we put everything in a combined stream with value and metadata and emit whenever metadata changes (e.g. 2/3 relays have the event) 
or
separate event metadata api that emits independently.
For direct consumption to UI a separate API could be beneficial, it heavily depends on how many updates we send in the stream.
We could still include limited metadata for the origin (cache network) in the data stream but treat it as immutable, incomplete metadata info (similar to how its sketched in this ADR).

frnandu:
Verdict: not merge-ready yet. Most comments addressed, but one critical semantic gap remains.

  local-first-reads.md:L67/L95: 🔴 bug: (null, relays) means “confirmed absence,” yet one unreachable relay is tolerated. One reachable empty relay cannot prove absence when another relay may hold newer value. Define required coverage: all selected relays, quorum, authoritative set, or explicit “best observed” semantics.

  local-first-reads.md:L89-L92: 🔴 bug: no relay reachable produces error, but timeout before any relay answers returns cached value. Same execution can satisfy both rules. Add deterministic outcome table for EOSE, timeout, zero responses, partial responses, cached value, and cache miss.

  Your points:

  - Per-relay provenance: not solved; explicitly scoped out. Acceptable for typed high-level reads, but cache event sources do not replace request-scoped provenance.

  - Raw queries/subscriptions: properly scoped out.
  - Combined provenance: consciously deferred; acceptable.
  - Equal cache/relay values: fixed.
  - Replacement ordering: fixed using NIP-01 ordering.
  - Nullable/loading state: fixed with T?.
  - Failure/completion semantics: only partially fixed; blockers above remain.
  - Subscription lifecycle: properly scoped out.

  Recommendation: request changes covering relay-coverage and terminal outcome matrix. After that, remove whitespace, rebase, obtain green checks—then accept ADR. No need demand implementation tests from documentation-only PR.

frnandu (changes requested):
see above

PR #703, opened 2026-08-08
Fix/fetched ranges recording race

nogringo:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved fetched-range tracking to reflect events actually delivered by the network.
  * Corrected range calculations for limited queries and responses capped by relays.
  * Preserved gaps for older events that were not fetched, with accurate oldest-event boundaries.

* **Tests**
  * Added integration coverage for limited queries and relay response limits.
  * Updated expectations to validate returned event counts and accurate fetched-range status.

### DIFF
--- a/packages/ndk/lib/domain_layer/usecases/requests/requests.dart
+++ b/packages/ndk/lib/domain_layer/usecases/requests/requests.dart
@@ -61,14 +61,14 @@ class Requests {
     required EventVerifier eventVerifier,
     required List<EventFilter> eventOutFilters,
     required Duration defaultQueryTimeout,
-  })  : _engine = networkEngine,
-        _relayManager = relayManager,
-        _cacheWrite = cacheWrite,
-        _cacheRead = cacheRead,
-        _globalState = globalState,
-        _eventVerifier = eventVerifier,
-        _eventOutFilters = eventOutFilters,
-        _defaultQueryTimeout = defaultQueryTimeout;
+  }) : _engine = networkEngine,
+       _relayManager = relayManager,
+       _cacheWrite = cacheWrite,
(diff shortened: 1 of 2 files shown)


packages/ndk/lib/domain_layer/usecases/requests/requests.dart
```diff
@@ -328,17 +328,27 @@ class Requests {
       writeToCache: request.cacheWrite,
     );
 
+    final networkEvents = <Nip01Event>[];
+    final trackedNetworkStream = _fetchedRanges == null
```
1-leo, line 334: not cleaned up, potential memory leak

1-leo (changes requested):

packages/ndk/lib/domain_layer/usecases/requests/requests.dart
1-leo: the `trackedNetworkStream` is still not closed or am I missing something?

1-leo (changes requested):

packages/ndk/lib/domain_layer/usecases/requests/requests.dart
nogringo: `trackedNetworkStream` is just `preparedNetworkStream.map(...)`, a derived stream, not a controller, so there is nothing to close on it. It completes automatically when `state.networkController` is closed (EOSE / timeout / closeSubscription), and its only subscription is the one in `StreamResponseCleaner`, which ends via `onDone` and then closes `state.controller`. Same lifecycle as `preparedNetworkStream` before this change.

1-leo (approved):

Issue #707, opened 2026-08-12
Fetched ranges depends on auth context


nogringo:
A relay may respond differently depending on who is connected on the websocket connection.
So fetched ranges depends on who is connected on the WebSocket connection.

Issue #665, opened 2026-06-15
authenticateAs for broadcast


nogringo:
### User Story

**As a** user
**I want to** be able to broadcast an event signed by x as y
**So that** relays that only accepts y as broadcaster can broadcast any event

PR #710, opened 2026-08-13
Feat/one identity per connection

nogringo:
# One authenticated identity per relay connection

A connection is identified by `(relay url, pubkey | null)` instead of by url
alone. The key says which identity a socket may ever assume, not that it is
already authenticated:

- an anonymous connection is bound to nobody and never answers a challenge
- a bound connection may only ever authenticate as the account it was opened for

A relay refusing a request with `auth-required` no longer triggers an AUTH on
the socket it refused: the request moves to a connection bound to the account,
and both attempts stay in the request state. Relays send their challenge
whenever they want, and some only once a request needs one, so a bound
connection sends first and answers when asked.

## Breaking

- `relayConnectivityChanges` emits `List<RelayConnectivity>` instead of a map
  keyed by url, which could only carry one connection per relay
- `GlobalState.relays` and `RequestState.requests` are keyed by
  `RelayConnectionKey`, `registerRelayRequest` takes a `connectionKey`
- `NdkConfig.eagerAuth` is deprecated and has no effect: an anonymous
  connection never authenticates and a bound one always does. This one is
  silent, nothing fails to compile.

## Deliberately left out

- broadcasts still authenticate in place, `BroadcastState` is keyed by url and
  cannot follow an event across two connections (see #665)
- fetched ranges still ignore the identity (see #707). The key must become the
  identity whose AUTH was accepted, not the connection's binding, otherwise a
  bound connection that never authenticated records anonymous data as if it had
  been fetched identified and the gap is never re-fetched.

## Testing

Two new ones pin what would regress silently: the anonymous
connection survives a retry instead of being promoted, and closing all
transports leaves no authenticated connection behind.

## New Features
- Added support for separate anonymous and authenticated connections to the same relay.
- Added public relay connection identity support.
- Connectivity updates now support connection lists and per-connection management.
- Added a default three-second authentication challenge timeout.

## Bug Fixes
- Improved authenticated request routing, retries, reconnection, cleanup, and relay state tracking.
- Improved subscription and query replay after connection failures.

## Tests
- Expanded coverage for authentication, reconnection, connection identity, and cleanup scenarios.

### DIFF
--- a/doc/concepts/nip42-auth.md
+++ b/doc/concepts/nip42-auth.md
@@ -1,35 +1,3 @@
 # NIP-42 Authentication
 
 NDK handles NIP-42 relay authentication automatically. When a relay requires authentication, NDK will sign and send AUTH events, then retry the original request.
-
-## Auth Strategies
-
-NDK supports two authentication strategies:
-
-### Lazy Auth (default)
-
-```dart
-final ndk = Ndk(NdkConfig(
-  eventVerifier: Bip340EventVerifier(),
-  cache: MemCacheManager(),
-  // eagerAuth: false (default)
-));
-```
-
-- AUTH is sent **only after** the relay responds with `auth-required`
-- More privacy-respecting: doesn't reveal identity until necessary
-- Flow: `REQ` → `CLOSED auth-required` → `AUTH` → `OK` → `REQ` (retry)
-
-### Eager Auth
-
-```dart
-final ndk = Ndk(NdkConfig(
-  eventVerifier: Bip340EventVerifier(),
-  cache: MemCacheManager(),
-  eagerAuth: true,
-));
-```
-
-- AUTH is sent **immediately** when the relay sends a challenge
-- Faster for relays that always require auth
-- Flow: `AUTH challenge` → `AUTH` → `OK` → `REQ`
--- a/packages/ndk/lib/config/request_defaults.dart
+++ b/packages/ndk/lib/config/request_defaults.dart
@@ -27,4 +27,9 @@ class RequestDefaults {
 
   /// default timeout for AUTH callbacks (how long to wait for AUTH OK)
   static const Duration DEFAULT_AUTH_CALLBACK_TIMEOUT = Duration(seconds: 30);
+
+  /// how long to wait for a NIP-42 challenge once something asked us to
+  /// authenticate; a relay that wants auth sends it with or right after the
+  /// refusal, so waiting longer only stalls the request
+  static const Duration DEFAULT_AUTH_CHALLENGE_TIMEOUT = Duration(seconds: 3);
 }
--- a/packages/ndk/lib/domain_layer/entities/global_state.dart
+++ b/packages/ndk/lib/domain_layer/entities/global_state.dart
@@ -1,5 +1,6 @@
 import 'broadcast_state.dart';
 import 'nip77_state.dart';
+import 'relay_connection_key.dart';
 import 'relay_connectivity.dart';
 import 'request_state.dart';
 
@@ -17,9 +18,9 @@ class GlobalState {
   final Map<String, BroadcastState> inFlightBroadcasts = {};
 
   /// touched relays by ndk - connected, connecting, disconnected
-  /// key: relay url/identifier
+  /// key: relay url and identity bound to the connection
   /// value: relay connectivity
-  Map<String, RelayConnectivity> relays = {};
+  Map<RelayConnectionKey, RelayConnectivity> relays = {};
 
   /// clean urls of relays that are blocked
   Set<String> blockedRelays = {};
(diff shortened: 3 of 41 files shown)


frnandu (changes requested):
1. Async authentication can resurrect a closed subscription
       packages/ndk/lib/domain_layer/usecases/relay_manager.dart:1184 and 1223
  
       If closeSubscription() runs while openConnectionAs() or authenticateConnection() is pending, the callback later resumes and sends REQ anyway. RequestState.close() does not clear state.requests, so the check at line 1225 still succeeds after removal from inFlightRequests.
  
       Result: an orphan relay subscription exists after closeSubscription() returns, and NDK no longer tracks it to send CLOSE.
  
       Fix: before each asynchronous retry, verify:
       identical(globalState.inFlightRequests[reqId], state)
       and ensure the state/controllers remain open. Ideally closing the request should cancel pending auth/open operations.
  
    2. Authentication state is not scoped to a transport generation
       packages/ndk/lib/domain_layer/usecases/relay_manager.dart:953-1035
  
       _challengeWaiters, _authenticating, and pending AUTH callbacks are keyed only by RelayConnectionKey and survive reset, close, or disconnect.
  
       Example:
       - Start AUTH while waiting for a challenge.
       - Close and reopen the same connection key.
       - The replacement socket’s challenge completes the old waiter.
       - _sendAuth() still holds the old RelayConnectivity, while the new attempt shares the stale _authenticating future.
  
       This can send AUTH to a closed transport, leave the new socket unauthenticated until timeout, or let an old AUTH response mark a replacement socket authenticated.
  
       Fix: add a transport-generation token and cancel/fail all authentication state whenever that transport closes or is replaced.

    3. closeConnection() does not forget transportless entries
       packages/ndk/lib/domain_layer/usecases/relay_manager.dart:1451-1459
  
       Removal and auth cleanup are conditional on relayTransport != null. After resetConnection() or a failed connection attempt, closeConnection() and closeAllTransports() leave the entry in globalState.relays.

       Remove and clean the entry whenever connectivity != null; guard only the actual transport close.

    4. Authenticated retries corrupt activeRequests
       packages/ndk/lib/domain_layer/usecases/relay_manager.dart:1237-1240

       The retry sent on the bound connection never increments that connection’s activeRequests, but EOSE or closeSubscription() decrements it through _sendCloseToRelay(). The authenticated connection can therefore report -1, while accounting is attached to the original anonymous attempt.

       Update request accounting during the handoff.

frnandu:
Remaining blocking findings
  
    5. A request retrying authentication can be closed while waiting for reconnection
       relay_manager.dart:1296-1301, 1483-1495
  
       When AUTH fails because its transport generation changed, retryingAuth is correctly retained. However, _checkNetworkClose() still treats every disconnected connection as finished at line 1491, regardless of retryingAuth.
  
       If the replacement socket is not open yet, the network controller closes and the request can be removed before reconnection replays it.
  
       Fix: a disconnected request with retryingAuth == true must not count as finished while reconnect/replay is pending.
  
    6. Terminal stream errors bypass auth cleanup and reconnection
       relay_manager.dart:678-689
  
       onError closes the connectivity but does not call _forgetAuthState() or initiate reconnection. Because closing cancels the stream subscription, onDone is not guaranteed to run afterward.
  
       Result: pending AUTH state can remain until timeout and active subscriptions may stay permanently disconnected despite allowReconnectRelays.
  
       Fix: route onError and onDone through one idempotent disconnect/cleanup/reconnect handler.

1-leo (commented):
Just for my understanding in the code there are multiple places in the jit engine with:
```dart
---      connectedRelays: relayManagerLight.connectedRelays
+++    connectedRelays: relayManagerLight.connectedAnonymousRelays
```
This implies that the jit engine is not using authenticated relays in the strategies, right?

The tests should tests both engines

nogringo:
The JIT engine uses authenticated connection if needed.

frnandu (changes requested):
packages/ndk/lib/domain_layer/usecases/relay_manager.dart:L1493: 🔴 bug (privacy): broadcast AUTH is sent on the shared RelayConnectionKey.anonymous socket. Once accepted, later “anonymous” requests reuse an authenticated connection and become attributable to the broadcast account. Reproduced with both engines. Open an identity-bound connection via openConnectionAs and retry the EVENT there.

  packages/ndk/lib/domain_layer/usecases/relay_manager.dart:L315: 🟡 risk: transient disconnects do not reset activeRequests; onReconnect replays every REQ and increments again, producing 1 → 2, then CLOSE → 1. Reset the transport-scoped count in onDisconnect before replay.

nogringo:
The privacy bug that you found on the broadcast path will be fixed in a following PR.
It was expected, I splited the work in multiple PR so it's easier to review small parts.
It will be fixed at the same time with #665

frnandu (approved):

PR #711, opened 2026-08-13
style: apply dart format for Dart 3.13

nogringo:
## Why

The latest stable Dart/Flutter release changed the formatter's output, so the `format` CI check went red.

## What

- `dart format --language-version=latest .` over the whole repo (137 files).
- Removed `final` from function parameters in `search.dart` and `contact_list.dart`. Dart 3.13 reserves `final` and `var` in parameter lists for primary constructors, so the formatter could not even parse those two files and silently skipped them.

### DIFF
--- a/packages/bc_ur/lib/bytewords.dart
+++ b/packages/bc_ur/lib/bytewords.dart
@@ -1,4 +1,5 @@
 import 'dart:typed_data';
+
 import 'package:ur/utils.dart';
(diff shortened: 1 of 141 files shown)


frnandu:
this would be good to do when there is not many open PR's otherwise the conflicts might be painfull to resolve....

1-leo (approved):

PR #712, opened 2026-08-14
fix!: replace GPL-3.0-only crystals-dilithium with fips204 (ML-DSA)

### LINKED ISSUE #717: GPL-3.0-only crystals-dilithium is statically linked into every build of the MIT-licensed ndk package
## Summary

`packages/ndk` is published under **MIT**, but its Rust backend depends on [`crystals-dilithium`](https://crates.io/crates/crystals-dilithium) `2.0.0`, which is **GPL-3.0-only**. The crate is compiled and statically linked into every build, including apps that never touch the quantum-secure signer.

## Evidence on `master`

- `packages/ndk/rust/Cargo.toml` declares `crystals-dilithium = "2.0.0"` as an unconditional dependency (no feature gate).
- crates.io reports the license for that exact version as `GPL-3.0-only`:
  ```
  curl -s https://crates.io/api/v1/crates/crystals-dilithium/2.0.0 | jq -r .version.license
  GPL-3.0-only
  ```
- `packages/ndk/rust/Cargo.toml` sets `crate-type = ["cdylib", "staticlib"]`, and `packages/ndk/hook/build.dart` runs `RustBuilder` unconditionally, so the native library is built for every dependent app.
- `packages/ndk/rust/src/lib.rs` imports `crystals_dilithium::{dilithium2, dilithium3, dilithium5}` at the top of the same file that holds the ordinary secp256k1 Schnorr verifier. There is no build configuration in which one is linked and the other is not.
- `LICENSE` at the repo root and `packages/ndk/LICENSE` are both MIT.

## Why this matters

Static linking of GPL-3.0-only object code makes the resulting binary a derivative work, which obliges downstream apps to distribute under GPL-3.0. Three consequences follow:

1. **The MIT promise on the package is not what consumers actually receive.** Someone adding `ndk` to a closed-source app inherits GPL-3.0 obligations they had no way to see from the pub.dev listing.
2. **It reaches consumers who never opted into post-quantum anything.** The quantum-secure signer is experimental, but the license effect is not opt-in, because the object code is linked either way.
3. **GPL-3.0 is widely held to be incompatible with App Store distribution terms**, so this can block iOS shipping for any app in the dependency tree.

`crystals-dilithium` is the only GPL-licensed crate in the tree. Everything else (`hex`, `secp256k1`, `sha2`, `serde_json`, `getrandom`) is MIT or Apache-2.0.

## Secondary problem in the same dependency

`crystals-dilithium` implements the round-3 CRYSTALS submission, not FIPS 204. NIST changed the algorithm during standardisation, so Dilithium and ML-DSA are different schemes with different wire formats. Keys and signatures produced by the current code can only be verified by the current code. No FIPS 204 implementation can read them, and vice versa. For a signature scheme, that is a defect rather than a preference, and it means there is no compatibility cost to replacing the crate now.

## Proposed fix

Replace it with [`fips204`](https://crates.io/crates/fips204), which is `MIT OR Apache-2.0` and standards-conformant. That is what #712 does, and it also adds a reference-vector test pinning an ML-DSA-87 public key byte-for-byte against `@noble/post-quantum`.

If you would rather not take that PR, the license conflict still needs resolving one way or another. The alternatives I can see:

- Feature-gate the quantum-secure code so `crystals-dilithium` is not linked by default (does not fix the interop problem, and still leaves a GPL path for anyone enabling the feature).
- Remove the quantum-secure signer from `packages/ndk` entirely and ship it as a separate, clearly GPL-3.0 package.
- Relicense `packages/ndk` as GPL-3.0 (almost certainly not what you want for an SDK).

Happy to adjust #712 in whatever direction you prefer.

leonacostaok:
Closes #717.

## The licensing problem

`packages/ndk/rust` depends on `crystals-dilithium`, which is **GPL-3.0-only**. NDK ships under **MIT**.

The crate is not optional at build time. `packages/ndk/rust/Cargo.toml` declares `crate-type = ["cdylib", "staticlib"]`, and `packages/ndk/hook/build.dart` compiles it unconditionally for every dependent app. The ordinary Schnorr verifier lives in the same `src/lib.rs` as the quantum-secure signer, so an app that never touches post-quantum code still links the GPL-3.0-only object code.

Statically linking GPL-3.0-only code makes the combined binary a derivative work, which obliges every downstream app to distribute under GPL-3.0. That is a hard problem for closed-source consumers, and GPL-3.0 is also widely held to be incompatible with App Store distribution terms.

This PR replaces it with [`fips204`](https://crates.io/crates/fips204), which is **MIT OR Apache-2.0** and imposes none of that.

## The interoperability problem

`crystals-dilithium` implements the round-3 CRYSTALS submission. NIST changed the algorithm during standardisation, so Dilithium and ML-DSA are different schemes with different wire formats.

Keys and signatures produced by the current code can therefore be verified only by the current code. No FIPS 204 implementation can read them, and vice versa. For a signature scheme whose entire purpose is letting someone else check your work, that is a defect rather than a preference.

## Also in this change

**Seed derivation.** Keys were random-only, so a signing key could never be restored from a mnemonic — losing it lost the identity permanently. `qs_derive_keypair_from_seed` derives from a 64-byte BIP-39 seed via HKDF-SHA256 with `nip-pqc/v1/ml-dsa-<level>/<account>`, making the key a *sibling* of the secp256k1 key rather than a child: breaking secp256k1 does not reach it, and one mnemonic restores both. A 32-byte secp256k1 private key is rejected as input, because deriving from it would be circular.

**Two safety fixes found while auditing this code:**

- `write_buffer` leaked a `Vec` recording only its length, while `qs_free_buffer` reconstructed it with `Vec::from_raw_parts(data, len, len)`. That is undefined behaviour whenever capacity exceeds length. It holds for every value passed today, so this was a latent trap rather than a live bug — the next contributor building an output with `push` or `extend` would have introduced heap corruption in the free path with no compiler diagnostic. `into_boxed_slice` reallocates to the exact size.
- `qs_free_buffer` now zeroizes before deallocating. These buffers carry secret keys, and a freed-but-unwiped secret is recoverable from a core dump or swap file.

## Breaking change

`level` now takes the ML-DSA parameter numbers (44/65/87) instead of the Dilithium ones (2/3/5). The old values are **rejected rather than remapped**, so an un-updated caller fails loudly instead of silently receiving different security properties than it asked for.

Existing quantum-secure keys do not carry over. They could never have interoperated with anything, and the signer is shipped as experimental.

## Verification

Run locally against Flutter 3.41.4 (the pinned `FLUTTER_VERSION`) and Rust 1.93.

**Rust**
- `cargo test` — **12/12 pass**, including `qs_mldsa87_matches_fips204_reference_vector`, which pins an ML-DSA-87 public key byte-for-byte against `@noble/post-quantum` from the same derived seed. That is precisely the interoperability the current implementation cannot satisfy.
- `cargo clippy --all-targets` — clean.
- `cargo fmt --check` — clean.

**Dart**
- `flutter analyze --fatal-infos --fatal-warnings` on `packages/ndk` — **No issues found.**
- `dart format` — `rust_lib.dart` is clean under Dart 3.12.2, matching the formatter your CI runs.
- The FFI was exercised end to end against the compiled library, not just analyzed. Deriving an ML-DSA-87 keypair from a 64-byte seed, signing, and verifying round-trips through `qs_derive_keypair_from_seed` / `qs_sign` / `qs_verify`, and the derived public key is 2592 bytes as FIPS 204 requires. The legacy Dilithium levels 2/3/5 are confirmed rejected rather than remapped, and a 32-byte secp256k1 private key is confirmed rejected as seed input.

The existing `test/scenarios/qs_sign_verify_test.dart` is `skip: true` on `master` and I left it that way; the checks above were run through a separate harness rather than by changing your test gating.

## Summary by CodeRabbit

* **New Features**
  * Added deterministic ML-DSA keypair derivation from a 64-byte BIP-39 seed and account number.
  * Native signing and verification now support ML-DSA parameter sets 44, 65, and 87.
* **Changes**
  * ML-DSA level 87 is now the default for signing, verification, and key generation.
  * Legacy Dilithium security levels are no longer accepted.
  * Updated terminology and documentation to align with FIPS 204.

### DIFF
--- a/packages/ndk/lib/data_layer/repositories/signers/qs_rust_event_signer.dart
+++ b/packages/ndk/lib/data_layer/repositories/signers/qs_rust_event_signer.dart
@@ -1,7 +1,7 @@
 /// Quantum-secure event signer with platform-specific implementations.
 ///
 /// On native platforms (Android, iOS, Linux, macOS, Windows), this uses FFI
-/// to call Rust code for CRYSTALS-Dilithium signing.
+/// to call Rust code for ML-DSA (FIPS 204) signing.
 ///
 /// On web platforms, this exports a stub that throws [UnsupportedError].
 library;
--- a/packages/ndk/lib/data_layer/repositories/signers/qs_rust_event_signer_native.dart
+++ b/packages/ndk/lib/data_layer/repositories/signers/qs_rust_event_signer_native.dart
@@ -9,7 +9,7 @@ import '../../../domain_layer/entities/pending_signer_request.dart';
 import '../../../domain_layer/repositories/event_signer.dart';
 import '../../../src/rust_lib.dart' as rust_lib;
 
-/// Holds a Dilithium keypair (public key + full keypair bytes for signing).
+/// Holds an ML-DSA keypair (public key + full keypair bytes for signing).
 class QsKeypair {
   final Uint8List publicKeyBytes;
   final Uint8List keypairBytes;
@@ -23,9 +23,9 @@ class QsKeypair {
 }
 
 /// An implementation of [EventSigner] that uses quantum-secure
-/// CRYSTALS-Dilithium signatures via native Rust FFI.
+/// ML-DSA (FIPS 204) signatures via native Rust FFI.
 class QsRustEventSigner implements EventSigner {
-  /// The Dilithium security level (2, 3, or 5).
+  /// The ML-DSA parameter set: 44, 65 or 87.
   final int level;
 
   late final QsKeypair _keypair;
@@ -36,7 +36,7 @@ class QsRustEventSigner implements EventSigner {
   /// Creates a [QsRustEventSigner] from an existing [QsKeypair].
   ///
   /// Use [QsRustEventSigner.generate] to create a new keypair first.
-  QsRustEventSigner({required QsKeypair keypair, this.level = 2})
+  QsRustEventSigner({required QsKeypair keypair, this.level = 87})
     : _keypair = keypair;
 
   @override
@@ -48,10 +48,11 @@ class QsRustEventSigner implements EventSigner {
   @override
   Iterable<String> get signerTransportRelayUrls => const <String>[];
 
-  /// Generates a new Dilithium keypair.
+  /// Generates a new ML-DSA keypair.
   /// Its only added here for testing purposes!
   ///
-  /// [level] selects the security level: 2 (~AES-128), 3 (~AES-192), or 5 (~AES-256).
+  /// [level] selects the ML-DSA parameter set: 44 (~AES-128), 65 (~AES-192), or
+  /// 87 (~AES-256, the CNSA 2.0 set). The old Dilithium values 2/3/5 are rejected.
   ///
   /// Returns a [QsKeypair] that can be stored and later passed to the constructor.
   ///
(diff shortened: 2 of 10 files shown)


leonacostaok:
Hey @frnandu & @nogringo, the license issue is real... Please, check it out!

1-leo:
Thank you @leonacostaok for your contribution!
One small thing I noticed you used not the latest version of `zeroize` and `hkdf` is there a particular reason for this?
Otherwise will upgrade them in a seperate pr

leonacostaok:
@1-leo no worries. thanks for the quick response for the community!

there's no specific reason, i didn't realize. please feel free to upgrade them. thanks!

PR #713, opened 2026-08-14
feat: add post-quantum hybrid encryption for direct messages (ML-KEM-1024)

leonacostaok:
> **Stacked on #712 — please merge that one first.**
>
> This branch is built on top of #712, so the diff below currently contains that PR's commit as well. Once #712 lands, this diff narrows on its own to just the commit added here. I would have based this PR on #712's branch directly, but a cross-fork PR can only target a branch that exists in this repository, so it targets `master`.
>
> The commit to review here is the second one, `feat: add post-quantum hybrid encryption for direct messages (ML-KEM-1024)`.

## Why

NDK can already sign post-quantum, but not encrypt. That leaves the more urgent half unaddressed.

NIP-44 derives its conversation key from a secp256k1 ECDH secret, so every encrypted event sitting on a public relay today becomes readable the day secp256k1 falls — **harvest now, decrypt later**. It is also the only half that can be fixed *in advance*: a message protected today stays confidential permanently, whenever the break arrives. Signatures cannot be pre-empted that way; they only help once the ecosystem stops accepting secp256k1. The two are complementary, and this adds the missing one.

## What

`pq.rs` exposing derivation, seal and open over the existing `QsBuffer` FFI convention, plus a `PqEncryption` Dart API and bindings.

**Design, and why:**

- **Hybrid, never bare.** The ML-KEM shared secret is combined with the classic NIP-44 conversation key through HKDF, so the result is no weaker than either input. Two tests assert this actually binds — a wrong conversation key fails and a wrong ML-KEM key fails. If either passed, the construction would not be hybrid.
- **Keys are siblings of the secp256k1 key, not children.** Deriving from the private key would be circular: an adversary who recovers it from the published pubkey repeats the derivation. A 64-byte BIP-39 seed is required, which rejects a 32-byte secp256k1 private key outright.
- **Its own version byte** rather than NIP-44's, so this can be adopted, renumbered or superseded without squatting a number in a registry it does not own.
- **The framing is authenticated.** Version, algorithm and both pubkeys go into the AEAD's associated data, so a ciphertext cannot be replayed into another conversation, have its direction swapped, or have its algorithm byte downgraded.
- **Length is padded** with NIP-44's scheme, so ciphertext size does not leak message size. Messages of 1, 2, 20 and 32 bytes produce identical wire sizes.
- **One generic failure.** Distinguishing bad padding from a bad tag from a wrong key would hand an attacker an oracle.
- **No relay changes.** The payload rides inside unchanged NIP-44 and NIP-59 gift wrap, so clients without support are unaffected.

Party pubkeys are validated as 64 lowercase hex characters. The associated data joins them with `:`, so an unvalidated string containing a colon would make distinct conversations produce identical associated data — sealing as `("aaaa:bbbb", "cccc")` would open as `("aaaa", "bbbb:cccc")`, defeating exactly the binding this advertises. Validation is wire-compatible: it rejects nothing a conforming implementation would send. It also turns a silently-unreadable uppercase-hex message into a loud failure at the boundary.

Secret material is zeroized: the derived KEM seed, the HKDF input keying material, the hybrid AEAD key, the encapsulation randomness and the decrypted padded buffer.

## Scope

This does **not** stop event forgery — events are still signed with secp256k1. It addresses confidentiality only. If ML-KEM fails, security falls back to today's NIP-44 rather than below it.

## Verification

Run locally against Flutter 3.41.4 (the pinned `FLUTTER_VERSION`) and Rust 1.93.

**Rust**
- `cargo test` — **32/32 pass** across both commits in this stack.
- Includes `opens_envelope_produced_by_typescript_reference`: a fixed vector sealed by [`@nostr-wot/pq`](https://www.npmjs.com/package/@nostr-wot/pq) and opened by this Rust code, so the two implementations are demonstrably wire-compatible rather than merely intended to be.
- Negative coverage: `wrong_conversation_key_fails`, `wrong_kem_key_fails`, `swapped_parties_fail`, `downgraded_alg_byte_fails`, `tampered_ciphertext_fails`, `colon_injection_in_party_is_rejected`, `padding_hides_message_length`.
- `cargo clippy --all-targets` and `cargo fmt --check` — clean.

**Dart**
- `flutter analyze --fatal-infos --fatal-warnings` on `packages/ndk` — **No issues found.**
- `dart format` — both touched files clean under Dart 3.12.2, matching the formatter your CI runs.
- `PqEncryption` was exercised end to end against the compiled library, not just analyzed:
  - a full **seal → open round trip** between two distinct parties, with the NIP-44 conversation key derived by real secp256k1 ECDH, recovering the exact plaintext;
  - **a wrong ML-KEM secret key fails to open**, confirming the hybrid binding holds through the Dart layer and not only in the Rust tests;
  - ML-KEM-1024 derivation returns a 1568-byte encapsulation key, is deterministic for a given seed, and is separated by account index;
  - a 32-byte secp256k1 private key is rejected as seed input;
  - `isPqEnvelope` correctly declines ordinary NIP-44 payloads and non-base64 input.

Two small hygiene fixes went in alongside the above, both consistent with the zeroization discipline described earlier: the conversation-key wipe now covers the key's actual length instead of a hardcoded 32, and the plaintext staging buffer is wiped before it is freed.

Happy to iterate on naming, placement, or the API shape — the Rust side is where the security properties live, and it is the part I can stand behind.


1-leo:
Hey, 
Thank you for your work, appreciate it!
Just as a reminder (you are probably aware) the QS stuff in NDK is marked as experimental and in its exploration phase, aiming to help the discussion in nostr with practical limitations.
We are generally advising to use this in production clients yet.

If you want to further contribute to ndk you are welcome to join our weekly dev calls https://dart-nostr.com/library-development/contributing/#communication-channels

1-leo:
@leonacostaok  fyi: 
I will add this to the sample app as well as the documentation (basically this pr description) If you want to shape that feel free to propose or edit something https://github.com/relaystr/ndk/issues/720

Issue #704, opened 2026-08-09
Requests: no way to tell an exhausted relay from a silent one


nogringo:
`NdkResponse` gives the events but not what each relay did with the request.
A client cannot tell these three apart:

- the relay sent EOSE and has nothing more to give,
- the relay never answered and the request timed out,
- the request was never sent, because no socket was open.

Proposal: expose the per relay outcome on `NdkResponse`, for instance

```dart
/// Resolves with the outcome of the request on each relay it was sent to.
Future<Map<String, RelayRequestOutcome>> get relayOutcomes;

enum RelayRequestOutcome { eose, closed, timedOut, notSent }
```

frnandu:
what is missing here to close it @nogringo ?

PR #721, opened 2026-08-18
feat: expose the outcome of a request on each relay

nogringo:
Closes part of #704.

`NdkResponse` gave the events but not what each relay did with the request, so a caller could not tell an exhausted relay from a silent one.

- `relayOutcomes`, keyed by relay url, readable at any time
- `relayOutcomesDone`, the same once the request is over
- a `RelayRequestOutcome` carries the reason the relay gave: the message of a CLOSED, for instance `blocked: you are not whitelisted`, which was parsed and then dropped

Outcome types: `pending`, `eose`, `closed`, `disconnected`, `timedOut`, `notSent`.

Notes:

- keyed by url, not by connection: several connections to one relay (an anonymous attempt handed over to an authenticated one) collapse into the first outcome in declaration order, so an auth handover that ends on an EOSE reports `eose`
- `pending` is what a live subscription reports, and what the auth retry reports while it is on its way back
- `notSent` is declared but nothing emits it yet: the engines drop a relay they could not reach. Two follow-ups, one per engine
- a paginated query merges its pages, last page wins per relay
- a request merged into an identical one in flight reports the outcomes of the request serving it

Tests run against both engines: EOSE, CLOSED with its message, timeout, live subscription, auth refused, auth handover collapse, merged duplicate.

## Summary by CodeRabbit

- **New Features**
  - Added relay request outcomes, including pending, completed, closed, disconnected, authenticated, and timed-out states.
  - Responses now provide current relay outcomes, live outcome updates, and a completion signal.
  - Paginated requests aggregate and stream relay outcomes across pages.
  - Relay closure messages are preserved and surfaced with the corresponding outcome.

- **Bug Fixes**
  - Requests now promptly report disconnected relays instead of waiting for timeout or completion events.


packages/ndk/lib/domain_layer/entities/request_response.dart
```diff
@@ -15,12 +16,43 @@ class NdkResponse {
...
+  /// not answered yet is [RelayRequestOutcomeType.pending], which is what a
+  /// live subscription shows for as long as it runs.
+  Map<String, RelayRequestOutcome> get relayOutcomes => _relayOutcomes();
+
+  /// [relayOutcomes] once the request is over, either because every relay
+  /// ended it or because it ran into its timeout.
+  ///
+  /// A subscription is only over once it is closed, so this resolves on
+  /// `closeSubscription` for one.
+  Future<Map<String, RelayRequestOutcome>> get relayOutcomesDone =>
```
1-leo, line 51: no stream interface?

packages/ndk/lib/domain_layer/entities/request_state.dart
```diff
@@ -71,6 +78,13 @@ class RequestState {
   /// timeout duration, closes all streams
   Duration? timeoutDuration;
 
+  /// set when the request ended on its timeout instead of on the relays
+  bool timedOut = false;
+
+  /// request this one was merged into by the concurrency check, when its stream
+  /// got replaced by an identical request already in flight
+  RequestState? servedBy;
```
1-leo, line 172: makes sense to link the original request

packages/ndk/lib/domain_layer/entities/request_state.dart
```diff
@@ -137,6 +152,49 @@ class RequestState {
         !element.retryingAuth,
   );
 
+  /// What the request ended with on each relay it was sent to, as it stands now
+  ///
+  /// Keyed by relay url: several connections to one relay collapse into the
+  /// outcome that comes first in [RelayRequestOutcomeType].
+  Map<String, RelayRequestOutcome> get relayOutcomes {
```
1-leo: may be better with a stream setup, probably even more efficient as we dont need the for loop to run every time

packages/ndk/lib/domain_layer/usecases/requests/requests.dart
```diff
@@ -506,7 +518,9 @@ class Requests {
             ),
           );
 
-          return MapEntry(relay, await response.future);
+          final pageEvents = await response.future;
+          relayOutcomes.addAll(response.relayOutcomes);
+          return MapEntry(relay, pageEvents);
```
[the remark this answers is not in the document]
1-leo: @nogringo we should check the pagination per relay, not sure how it is handled now

1-leo (commented):
need to discuss how if/how we do a streaming interface for the relay request state reporting

packages/ndk/lib/domain_layer/usecases/requests/requests.dart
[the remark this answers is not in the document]
nogringo: It was pre-existent so I fixed it in https://github.com/relaystr/ndk/pull/764

packages/ndk/lib/domain_layer/entities/request_state.dart
```diff
@@ -137,11 +238,138 @@ class RequestState {
...
+    }
+    if (_relayOutcomes[url] == collapsed) {
+      return false;
+    }
+    _relayOutcomes[url] = collapsed;
+    return true;
+  }
+
+  void _publishRelayOutcomes() {
+    _relayOutcomesSnapshot = Map.unmodifiable(_relayOutcomes);
```
frnandu: L315: risk: each relay registration/completion rescans requests and copies the entire outcome map even when nobody reads outcomes, making ordinary R-relay fan-out O(R^2). Keep raw state updates O(1), derive on read, and publish snapshots only while the outcome stream has listeners.

packages/ndk/lib/domain_layer/usecases/relay_manager.dart
```diff
@@ -1531,6 +1530,12 @@ class RelayManager<T> {
     );
 
     if (didAllRelaysFinish) {
+      for (final key in myNotConnectedRelays) {
+        final request = state.requests[key]!;
+        if (!request.receivedEOSE && !request.receivedClosed) {
+          request.connectionGone = true;
```
frnandu, line 1557: L1536: bug: this marks disconnections only when `_checkNetworkClose` happens to run; `_handleTransportGone` never invokes it for ordinary requests after reconnect fails or is disabled, leaving subscriptions pending forever and queries reported as timed out. Mark matching in-flight requests `connectionGone` on terminal reconnect failure.

packages/ndk/lib/domain_layer/usecases/relay_sets_engine.dart
```diff
@@ -282,7 +282,7 @@ class RelaySetsEngine implements NetworkEngine {
         in state.requests.entries) {
```
frnandu, line 282: L282: bug: removing a failed send makes public `RelayRequestOutcomeType.notSent` impossible, although issue #704 requires callers to distinguish it. Retain a terminal send-failure/message state and return `notSent`, or remove that outcome from this PR.

packages/ndk/lib/domain_layer/usecases/requests/requests.dart
```diff
@@ -489,6 +510,7 @@ class Requests {
       );
 
       final initialEvents = await initialResponse.future;
+      mergeRelayOutcomes(initialResponse.relayOutcomes);
```
frnandu: L513: bug: child outcomes are read only after `future` completes, so paginated `relayOutcomesStream` emits no pending or intermediate state despite its contract. Forward each child `relayOutcomesStream` while its page is active and retain the final snapshot.

packages/ndk/lib/domain_layer/usecases/requests/requests.dart
```diff
@@ -563,7 +585,9 @@ class Requests {
             ),
           );
 
-          return MapEntry(relay, await response.future);
+          final pageEvents = await response.future;
+          mergeRelayOutcomes(response.relayOutcomes);
```
frnandu: L589: bug: this response may cover the whole `relaySet` because the page request retains `relaySet`, which overrides `explicitRelays: [relay]`; `addAll` then overwrites unrelated outcomes in completion order and repeats network work. Clear `relaySet` on subsequent pages, or merge #764 first.

packages/ndk/lib/domain_layer/entities/request_response.dart
```diff
@@ -15,12 +16,58 @@ class NdkResponse {
...
+  })  : _relayOutcomes = relayOutcomes ?? _noOutcomes,
+        _relayOutcomesStream = relayOutcomesStream ?? _noOutcomesStream,
+        _relayOutcomesDone = relayOutcomesDone ??
+            Future.value(const <String, RelayRequestOutcome>{});
+
+  static Map<String, RelayRequestOutcome> _noOutcomes() =>
+      const <String, RelayRequestOutcome>{};
+
+  static Stream<Map<String, RelayRequestOutcome>> _noOutcomesStream() =>
+      const Stream<Map<String, RelayRequestOutcome>>.empty();
```
frnandu: L70: risk: the fallback closes without emitting the initial empty snapshot promised by `relayOutcomesStream`. Return `Stream.value(const <String, RelayRequestOutcome>{})` and cover the fallback.

frnandu (changes requested):
Requesting changes for outcome correctness and hot-path performance. Inline comments cover the PR-relevant blockers. CodeRabbit's deduplication-key and per-filter fetched-range findings are valid pre-existing issues and should be tracked separately instead of expanding this PR.

packages/ndk/lib/domain_layer/entities/relay_request_outcome.dart
```diff
@@ -0,0 +1,52 @@
+/// What a relay did with a request.
+///
+/// Declared in collapse precedence order: one request can run on several
+/// connections to the same relay, an anonymous one handed over to an
+/// authenticated one, and the url reports the first of these its connections
+/// reached.
+enum RelayRequestOutcomeType {
```
frnandu: L7: naming: keep `RelayRequestOutcome` for the value object, rename `RelayRequestOutcomeType` to `RelayRequestStatus`, and rename its `type` field to `status`. `pending` makes `Result` misleadingly terminal, while `outcome.status` stays clear for both live and completed requests.

packages/ndk/lib/domain_layer/usecases/relay_manager.dart
nogringo: Fixed. `_handleTransportGone` now gives up on what the connection still owed us, once the reconnect failed or was never tried, instead of waiting for a `_checkNetworkClose` nothing triggers. Tested on `RelayManager` with the plain transport: the one `Ndk` uses reconnects on its own and keeps its message stream open, so no socket ever dies for good under it.

packages/ndk/lib/domain_layer/usecases/relay_sets_engine.dart
nogringo: This is why this PR only close part of #704, I will do a following PR for that.

packages/ndk/lib/domain_layer/usecases/requests/requests.dart
nogringo: Fixed. Each page is followed through its `relayOutcomesStream` while it runs, with the final merge kept on completion.

packages/ndk/lib/domain_layer/usecases/requests/requests.dart
nogringo: We will merge #764 first.

1-leo (approved):
valid points from @frnandu otherwise good

packages/ndk/lib/domain_layer/entities/relay_request_outcome.dart
nogringo: Done

packages/ndk/lib/domain_layer/entities/request_state.dart
nogringo: Fixed, minus one part. The connections are now indexed by url, so collapsing a relay reads its own connections instead of scanning every connection of the request, and the handed-out snapshot is built on demand: a change only invalidates it, so a request nobody reads or watches copies nothing. That takes the R relay fan-out down from R maps allocated and R scans to neither.

I kept deriving on the change rather than on the read, though. `relayOutcomes` is readable at any time, so a caller polling it would pay the collapse on every read, and the recompute only runs on registration, EOSE, CLOSED and disconnection, never per event.

packages/ndk/lib/domain_layer/entities/request_response.dart
nogringo: Done

frnandu:
notSent unreachable:               Not fixed

  Blocking finding:

  - relay_request_outcome.dart: RelayRequestStatus.notSent remains public but impossible to observe. Failed sends remove request state, while  _outcomeOf never returns notSent. Either preserve failed-send state and emit notSent, or remove enum value until follow-up.

frnandu (changes requested):

frnandu (approved):

PR #724, opened 2026-08-18
feat: wallet improvements

### LINKED ISSUE #741: feat: bolt12 offer receive-only wallet

### LINKED ISSUE #811: Cashu wallet backup is clipboard-only and silently truncates on Android, while reporting success
## Summary

The Cashu wallet backup is offered only as "copy to clipboard". On Android the
clipboard silently truncates large payloads, so a big wallet produces a
**partial backup while the UI reports success**. The user then believes they
hold a valid backup of the only data that can recover their funds.

This was hit in production by a user of an SK Bitblik coordinator whose
transaction backup was roughly 20 000 lines of JSON. He only noticed the
truncation by chance, after comparing what he pasted with what he expected.

## Where

`packages/ndk_flutter/lib/widgets/wallets/wallet_action_dialogs.dart:284`

```dart
onPressed: () async {
  await Clipboard.setData(ClipboardData(text: backupJson!));
  displaySuccess(l10n.backupCopiedToClipboard);
},
```

The success message fires unconditionally. `Clipboard.setData` resolves whether
or not the platform kept the whole string, so there is nothing here that could
notice a truncated copy.

There is no file export anywhere in `widgets/wallets/` — no `writeAsString`, no
share sheet, no file picker. Clipboard is the only backup path, and
`showRestoreDialog` takes the backup back in by paste, so the same ceiling
applies in both directions.

## Why this matters

This is the recovery path for a wallet that, per #809, can already strand funds
permanently when local data is lost. A backup feature that silently yields an
incomplete file turns "you had a backup" into "you thought you had a backup".
The two bugs compose badly: #809 makes local data the only thing standing
between the user and unrecoverable sats, and this one quietly corrupts the
copy of it.

## Suggested fix

1. **Export to a file** — a share sheet or file picker — as the primary path.
   This removes the size ceiling entirely and is what a backup should be.
2. If the clipboard is kept as a convenience, **verify it**: read back with
   `Clipboard.getData` and compare length/hash against what was written, and
   show a failure instead of `backupCopiedToClipboard` when they differ.
3. Show the backup's size in the dialog so a truncated paste is visible.
4. Accept a file in `showRestoreDialog` too, otherwise a large backup can be
   written but never restored.

## Workaround that worked in the field

The affected user connected the phone to an external monitor and switched to
Android desktop mode, where the clipboard behaved differently and carried the
full JSON. He then extracted the stranded mint quote from that backup and
completed the mint with a script. Funds recovered — but nobody should need a
monitor and a custom script to read their own backup.

### LINKED ISSUE #815: Cashu wallet mint discover with ratings from nostr nip-87

### LINKED ISSUE #816: LNBits wallet integration

frnandu:
implements https://github.com/nostr-wallet-connect/nwc/pull/2

fixes #811
fixes #741
fixes #815 
fixes #816 

## New Features
- Added support for receive-only BOLT12 wallets, including offer, BIP-321 URI, BIP-353 address, and QR-code input.
- Added wallet-to-wallet transfers with compatible payment protocol selection.
- Added BIP-321 payment and receiving support across supported wallet connections.
- Added transaction details, payment proofs, configurable amounts, notes, descriptions, metadata, fees, and timeouts.
- Added localized wallet transfer and BOLT12 setup messaging.

## Bug Fixes
- Improved validation for invoices, offers, amounts, descriptions, payment instructions, and unsupported options.

## Tests
- Added coverage for BOLT12 wallets, transfers, BIP-321, NWC operations, permissions, and error handling.


packages/ndk/pubspec.yaml
```diff
@@ -24,6 +24,8 @@ platforms:
   windows:
 
 dependencies:
+  dart_bip353: ^0.8.0
+  dart_bolt12_decoder: ^0.8.0
```
nogringo: I would prefer if we do not add those two dependencies.

nogringo (changes requested):

packages/ndk/pubspec.yaml
frnandu: Why not?

nogringo: `ndk` is a published library, so every dependency here is imposed on all consumers.

**`dart_bolt12_decoder`** (measured on the two offers in `bolt12_wallet_test.dart`):

- The only key it adds over the in-repo decoder is `offer_id`.
- `offerId` is never read: storage, copy constructors, back to storage.
- Its `offer_id` is not BOLT12's merkle root, it hashes the same three TLV elements on every loop iteration.
- Returns `null` for offers with `offer_paths`, so it is already bypassed for modern offers.

Removing it is a net deletion and collapses `_validate` to a single code path.

**`dart_bip353`** (one call site, `_resolveBip353`):

- It is a DoH GET plus a string parse, on `http` which we already depend on.
- It checks `Status` but never the `AD` flag, so an unsigned domain is accepted as if signed. BIP353 requires a DNSSEC-signed record.
- The Cloudflare endpoint is hardcoded and not overridable.

Replacing it is roughly 40 lines and gets us the `AD` check plus a configurable endpoint.

packages/ndk/lib/domain_layer/entities/wallet/bip321.dart
```diff
@@ -0,0 +1,92 @@
...
+  /// Returns the BOLT11 amount in millisatoshis, or null when amountless.
+  static int? getBolt11AmountMsat(String invoice) {
+    final separator = invoice.toLowerCase().lastIndexOf('1');
+    if (separator < 0) {
+      throw const FormatException('Invalid BOLT11 invoice');
+    }
+
+    final hrp = invoice.toLowerCase().substring(0, separator);
+    final match = RegExp(
+      r'^ln(?:bcrt|bc|tb|sb)([0-9]*)([munp]?)$',
```
nogringo: Is `tbs` missing in the regex ?

nogringo (approved):
bolt12_wallet_provider.dart:328 : `amountSats` is a required parameter which is never read.

PR #726, opened 2026-08-19
fix: repeated signature verification

### LINKED ISSUE #715: Events already verified get their signature checked again on every delivery
`VerifyEventStream` verifies every event coming off the network unconditionally. Nothing consults the cache or the persisted `validSig` first, and no verifier memoizes (bip340, both Rust ones, `NdkEventVerifier`).

Deduplication happens only at the end of the pipeline, so a duplicate costs a Schnorr verification, a cache write and a cache read before being discarded. `returnedIds` is per request, so two subscriptions receiving the same event verify it twice.

nogringo:
Adds failing tests for #715: VerifyEventStream calls the verifier on every delivery, so a duplicate event is re-verified instead of being deduplicated first.

Covered cases:
- same event delivered several times on one stream
- event already carrying validSig true
- one event served by two relays in a single query
- two successive requests for the same event

fixes #715

## Summary by CodeRabbit

* **Improvements**
  * Reduced redundant signature verification for repeated, relayed, and concurrent event requests.
  * Verification results now account for exact event signatures and safely handle invalid or tampered events.
  * Duplicate event saves are prevented while preserving the original event content across cache implementations.
  * Added bounded cache eviction and cache clearing during shutdown.
  * Ephemeral events remain non-persistent.

* **Tests**
  * Added coverage for concurrent verification, duplicate persistence, relay deduplication, cache eviction, invalid signatures, and forged replacements.


packages/ndk/lib/domain_layer/usecases/requests/verify_event_stream.dart
```diff
@@ -9,35 +10,55 @@ class VerifyEventStream {
...
+    }
+    if (_cache.contains(data.id)) {
+      return Future.value(data.copyWith(validSig: true));
+    }
+
+    // mark eagerly (before awaiting) so duplicates delivered concurrently
+    // with this one also skip the verifier instead of racing it (the actual event is not!! marked verified == true)
+    _cache.markVerified(data.id);
+
+    return _verifyEvent(data);
```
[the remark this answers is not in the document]
frnandu: !!!! @1-leo this sounds critically important

frnandu (dismissed):
There is a critical correctness flaw in the optimistic cache approach: a concurrent duplicate is returned with validSig: true before the first verification finishes. If the first verification later fails—or throws—the duplicate has already escaped as trusted. I’m checking existing error tests and will reproduce this path directly.

packages/ndk/lib/domain_layer/usecases/requests/verified_event_cache.dart
```diff
@@ -0,0 +1,70 @@
...
+/// not commit to its signature. Callers must validate the incoming event id
+/// from its payload before consulting [hasVerifiedSignature].
+class VerifiedEventMemCache {
+  final int maxSize;
+  final LinkedHashSet<_VerificationKey> _verified =
+      LinkedHashSet<_VerificationKey>();
+  final Map<_VerificationKey, Future<bool>> _verificationInFlight =
+      HashMap<_VerificationKey, Future<bool>>();
+
+  VerifiedEventMemCache({this.maxSize = 20000});
```
nogringo: Reduce to 5000 ?

packages/ndk/lib/domain_layer/usecases/requests/verified_event_cache.dart
```diff
@@ -0,0 +1,70 @@
...
+typedef _VerificationKey = ({String id, String? signature});
+
+/// Bounded, in-memory LRU record of verified event id/signature pairs.
+///
+/// A matching id alone is not sufficient for reuse because an event id does
+/// not commit to its signature. Callers must validate the incoming event id
+/// from its payload before consulting [hasVerifiedSignature].
+class VerifiedEventMemCache {
+  final int maxSize;
+  final LinkedHashSet<_VerificationKey> _verified =
```
nogringo, line 16: Clear this on ndk destroy ?

PR #758, opened 2026-08-31
fix: multiple instances with dedicated GlobalState

### LINKED ISSUE #757: Multiple Ndk instances share static GlobalState and interfere with each other
**Describe the bug**

`Ndk` supports constructing more than one object, but all objects in the same Dart isolate share one static `GlobalState`:

```dart
class Ndk {
  static final GlobalState _globalState = GlobalState();

  Ndk(this.config)
      : _initialization = Initialization(
          ndkConfig: config,
          globalState: _globalState,
        );
}
```

`GlobalState` contains all in-flight requests, broadcasts, relay connections, blocked relays, and NIP-77 negotiations. Accounts, caches, engines, and `RelayManager` objects are otherwise created per `Ndk` instance. This creates a mixed ownership model: instance-local managers operate on process-global transport/request state.

As a result, two independent NDK contexts cannot safely coexist in one Flutter/Dart application. A practical example is an application with a permanent local trading identity and a separately authenticated administration/coordinator identity using NIP-55 or NIP-07.

Concrete consequences visible from the current implementation:

- `Ndk.destroy()` on either instance calls `closeAllSubscription()` and `closeAllTransports()` over the shared maps, closing the other instance's subscriptions and sockets.
- Anonymous relay sockets are shared by connection key. The `RelayManager` that originally created the socket owns its listener and `_accounts`, even when a request in shared state belongs to the other NDK instance.
- NIP-42 authentication can therefore be handled by a `RelayManager` whose account map does not contain the requesting account. Broadcast authentication may also fall back to that manager's logged account instead of the event author's account.
- Request concurrency/deduplication is shared across instances. Its hash covers filters, lifetime, and relays, but not the NDK instance, cache, verifier, output filters, or `authenticateAs` identity. Equivalent queries from independent contexts can consequently share a stream despite different identities/configuration.
- In-flight broadcasts are globally keyed by event ID, so the same event broadcast through two instances can replace the first instance's tracking state.
- Blocking a relay in one instance blocks it in every instance.
- Reconnection from one `RelayManager` replays every request in the global map, including requests owned by other instances.

**To Reproduce**

The lifecycle problem can be covered with a small test using two independent NDK objects:

```dart
final trader = Ndk(NdkConfig(
  cache: MemCacheManager(),
  eventVerifier: Bip340EventVerifier(),
  bootstrapRelays: [relayUrl],
));

final coordinator = Ndk(NdkConfig(
  cache: MemCacheManager(),
  eventVerifier: Bip340EventVerifier(),
  bootstrapRelays: [relayUrl],
));

final traderResponse = trader.requests.subscription(
  id: 'trader-subscription',
  filter: Filter(kinds: [1]),
  explicitRelays: [relayUrl],
);

// This currently walks the static request and relay maps.
await coordinator.destroy();

// The trader subscription/transport has also been closed and no longer
// receives events, even though trader.destroy() was never called.
```

Additional reproductions should use identical `cacheRead: true` queries on both instances with different `authenticateAs` accounts, and a NIP-42 relay where the anonymous socket was opened by the other instance.

**Expected behavior**

Each `Ndk` object should own its relay connections, request state, broadcast state, blocked-relay set, and NIP-77 state. Destroying, reconnecting, blocking a relay, querying, or broadcasting through one instance must not affect another instance.

If shared process state is intentionally supported as an optimization, it should be an explicitly injected/shared runtime whose lifecycle and account routing are designed for multiple owners, rather than an unconditional static singleton.

**Suggested implementation**

Make `GlobalState` instance-owned by default, for example through a private constructor that receives a newly allocated state:

```dart
factory Ndk(NdkConfig config) => Ndk._(config, GlobalState());

Ndk._(this.config, GlobalState state)
    : _globalState = state,
      _initialization = Initialization(
        ndkConfig: config,
        globalState: state,
      );
```

An optional explicitly shared runtime could be introduced separately if socket sharing between NDK contexts is a desired feature.

Suggested regression coverage:

1. Two instances can subscribe to the same relay/filter independently.
2. Destroying one instance leaves the other's subscription and connection alive.
3. NIP-42 requests authenticate with the account belonging to the requesting instance.
4. Broadcast acknowledgements and timeouts cannot cross instances.
5. Reconnection only replays subscriptions owned by that instance.
6. Relay blocking is instance-local.
7. Identical authenticated queries from different instances are not deduplicated together.

**Environment (please complete the following information):**

- NDK commit/version: `7f4cef7f6208cf39fc7946d62fde49707e811cec` (repository HEAD when reported)
- Runtime: Dart/Flutter, same Dart isolate
- Platforms relevant to the use case: Android, web, Linux
- Caches: independently constructed cache managers

**Additional context**

This is not primarily a request to reduce duplicated WebSocket connections. Independent sockets are acceptable and preferable when the instances represent security-sensitive identities. The problem is that the current static state shares sockets and request bookkeeping while accounts and managers remain instance-local, so the instances are neither isolated nor consistently coordinated.

frnandu:
fixes #757

## Summary by CodeRabbit

- **Bug Fixes**
  - Isolated relay, request, broadcast, and authentication state between separate NDK instances.
  - Improved relay discovery, connectivity validation, and fallback handling for author- and tag-based requests.
  - Fixed authentication-required broadcasts to retry reliably and resend events correctly.
  - Prevented duplicate in-flight NIP-05 requests from being shared across instances.
  - Improved concurrent relay connection handling for targeted requests.

- **Tests**
  - Added coverage for independent multi-instance behavior, concurrent authenticated broadcasts, relay connection timing, and fallback request handling.

### DIFF
--- a/packages/ndk/lib/domain_layer/usecases/relay_manager.dart
+++ b/packages/ndk/lib/domain_layer/usecases/relay_manager.dart
@@ -1401,24 +1401,14 @@ class RelayManager<T> {
     return logged != null && logged.signer.canSign() ? logged : null;
   }
 
-  /// Handles OK auth-required for broadcasts by authenticating and re-sending the EVENT
-  ///
-  /// This is the last place that still authenticates in place, because
-  /// BroadcastState is keyed by relay url and cannot yet track an event across
-  /// two connections. It goes away with the broadcast lot.
+  /// Handles OK auth-required for broadcasts by moving the retry onto an
+  /// account-bound connection, authenticating it once, and re-sending EVENT.
+  /// Concurrent broadcasts share [authenticateConnection], avoiding duplicate
+  /// AUTH events whose identical ids used to overwrite each other's callbacks.
   void _handleBroadcastAuthRequired(
     String eventId,
     RelayConnectivity relayConnectivity,
   ) {
-    final challenge = _lastChallengePerConnection[relayConnectivity.key];
-    if (challenge == null) {
-      Logger.log.w(
-        () =>
-            "Received OK auth-required but no challenge stored for ${relayConnectivity.url}",
-      );
-      return;
-    }
-
     final broadcastState = globalState.inFlightBroadcasts[eventId];
     if (broadcastState == null) {
       Logger.log.w(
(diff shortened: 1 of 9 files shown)


packages/ndk/lib/domain_layer/usecases/relay_manager.dart
```diff
@@ -1455,54 +1445,42 @@ class RelayManager<T> {
...
-      send(
-        relayConnectivity,
-        ClientMsg(ClientMsgType.kAuth, event: signedAuth),
-      );
-      Logger.log.d(
-        () =>
-            "AUTH sent for ${account!.pubkey.substring(0, 8)} to ${relayConnectivity.url}, waiting for OK...",
-      );
-    });
+    unawaited(retryOnBoundConnection());
```
nogringo, line 1521: Without `catchError`: an Exception in `openConnectionAs` or `send` becomes an unadled async error.

nogringo (commented):
### Relay candidates are now connected one at a time (relay_jit_pubkey_strategy.dart:210)

The old `.then()` started every `connectRelay` at once. With the `await` inside the loop, each candidate waits for the previous one, and an unreachable relay burns the full `DEFAULT_WEB_SOCKET_CONNECT_TIMEOUT` (4s) before the next one is even tried. On a follows feed the ranking easily holds 10+ relays.

### `_handleBroadcastAuthRequired` can fail a different broadcast (relay_manager.dart:1459 and :1475)

```dart
if (bound == null ||
    globalState.inFlightBroadcasts[eventId] != broadcastState) {
  failBroadcast(eventId, relayConnectivity.url, 'auth connection failed');
```

When that condition is true because the state was replaced (the same event broadcast a second time), `failBroadcast` resolves the id again and fails the **new** `BroadcastState`. Worth splitting the two cases: report the failure only while it is still the same state, otherwise just return.

nogringo:
There was another bug found during the review, but it is already fixed inside this pull request https://github.com/relaystr/ndk/pull/766

Both PRs touch the same files so it will be a merge conflict.

1-leo (approved):

PR #789, opened 2026-09-06
Fix/query without relays waits timeout

nogringo:
A query sent to zero relays only ended when its timeout fired. With no relay to answer, nothing closed the network stream: no EOSE, no CLOSED, no send failure. This happens whenever `bootstrapRelays` is empty and no `explicitRelays` are given, including when a malformed URL is silently dropped by `cleanRelayUrl`.

Worth a look: two tests in `query_timeout_test.dart` used `bootstrapRelays: ["invalid"]`, an URL `cleanRelayUrl` drops, so they were timing out on zero relays and pinned this very bug. They now point at a `MockRelay` that accepts the REQ and never answers (new `ignoreRequests` flag).

## Summary by CodeRabbit

* **Bug Fixes**
  * Requests with no available relays now finish promptly with an empty result instead of waiting for the full query timeout.
  * Relay discovery and connection handling now complete before requests are evaluated, improving reliability when using dynamic relay selection.
  * Requests are no longer dispatched when no relays are available.

* **Tests**
  * Added coverage for no-relay queries and nonresponsive relays to verify timeout behavior and prompt completion.


frnandu (approved):

packages/ndk/lib/domain_layer/usecases/jit_engine/relay_jit_request_strategies/relay_jit_pubkey_strategy.dart
```diff
@@ -272,6 +276,8 @@ class RelayJitPubkeyStrategy with Logger {
         );
       }
     }
+
+    await Future.wait(connectFutures);
```
1-leo: waiting will slow down the discovery process, it was left out on purpose.
tests pass without this line, why was it added?

1-leo (commented):
unsure about the await in jit engine.
Before it was required to not await for multiple filters, since we deprecated multiple filters its fine now

packages/ndk/lib/domain_layer/usecases/jit_engine/relay_jit_request_strategies/relay_jit_pubkey_strategy.dart
nogringo: It was there because `closeIfNoRelays()` must not fire while a send path is still resolving its connection, otherwise it sees an empty requests map and closes the stream.

The merged version does not wait at all: it checks `pendingConnections`, the counter master added for exactly this case, and `_blastTo` / `_sendTo` increment it synchronously before their first await. Nothing is slowed down.

1-leo (approved):

PR #803, opened 2026-09-10
feat: nip-82 applications releases zapstore support

### LINKED ISSUE #67: NIP-82 Applications/releases (zapstore support)
https://github.com/nostr-protocol/nips/pull/1336

frnandu:
fixes #67 

## Summary by CodeRabbit

- **New Features**
  - Added Nostr software discovery, release tracking, asset resolution, and compatible Android update selection.
  - Added Flutter update workflows, reusable update widgets, persistent release watching, and release engagement with publisher details, reactions, zaps, and comments.
  - Added an “installed version ahead of published release” status with multilingual support.
  - Added sample-app demonstrations and Android installation support.
- **Bug Fixes**
  - Improved download cancellation, validation, redirect handling, and cleanup.
  - Improved relay delivery handling for permanent policy failures.
- **Documentation**
  - Added Android app update publishing and integration guides.
- **Tests**
  - Added coverage for parsing, discovery, asset resolution, and update selection.


packages/ndk_flutter/android/src/main/kotlin/relaystr/ndk/DartNdkPlugin.kt
```diff
@@ -42,14 +54,328 @@ class DartNdkPlugin : FlutterPlugin, MethodCallHandler, ActivityAware,
...
+                _context.packageName,
+                android.content.pm.PackageManager.PackageInfoFlags.of(
+                    android.content.pm.PackageManager.GET_SIGNING_CERTIFICATES.toLong()
+                ),
+            )
+        } else {
+            @Suppress("DEPRECATION")
+            _context.packageManager.getPackageInfo(
+                _context.packageName,
+                android.content.pm.PackageManager.GET_SIGNING_CERTIFICATES,
```
nogringo: GET_SIGNING_CERTIFICATES does not populate signatures below API 28
This plugin declares minSdk = 21, so API 21 to 27 is in scope. There certificateHashes is always an empty list, and since _hasMatchingCertificate in software.dart is an any() over the installed hashes, every asset is filtered out. No update is ever offered on Android 8.1 and older. It fails closed, which is the safe direction, but it fails silently: there is no log or error, the update check simply finds nothing.

packages/ndk_flutter/lib/widgets/app_update/n_app_update_controller.dart
```diff
@@ -0,0 +1,519 @@
...
+    }
+  }
+
+  void _ensureReleaseSubscription() {
+    if (_releaseSubscription != null) return;
+    _releaseSubscription = ndkFlutter.ndk.software
+        .watchReleases(app: app, channel: channel, relays: relays)
+        .listen(
+          _evaluate,
+          onError: (Object error) {
```
nogringo, line 213: onError drops the subscription reference without cancelling it

packages/ndk/lib/domain_layer/usecases/software/software.dart
```diff
@@ -0,0 +1,297 @@
...
+              if (previous == null ||
+                  previous.event.createdAt < release.event.createdAt) {
+                byCoordinate[coordinate] = release;
+              }
+            }
+            final values = byCoordinate.values.toList()
+              ..sort(
+                (a, b) => b.event.createdAt.compareTo(a.event.createdAt),
+              );
+            controller.add(values);
```
nogringo: Every single incoming event produces an emission, including events that fail to parse, events rejected for a wrong author, identifier, or channel, and events whose coordinate is already held by a newer entry. In all of those cases the emitted list is byte for byte the one emitted just before.

packages/ndk_flutter/android/src/main/kotlin/relaystr/ndk/DartNdkPlugin.kt
```diff
@@ -42,14 +54,328 @@ class DartNdkPlugin : FlutterPlugin, MethodCallHandler, ActivityAware,
...
+                result,
+                File(directory, "update-${System.nanoTime()}.apk"),
+            ).also { _activeUpdate = it }
+        }
+
+        Thread {
+            try {
+                val apk = operation.apk
+                val connection = openUpdateConnection(url, operation)
+                val total = connection.contentLengthLong
```
nogringo: contentLengthLong requires API 24, plugin declares minSdk = 21

URLConnection.getContentLengthLong() was added in API 24. This module sets minSdk = 21, so on API 21 to 23 the call throws java.lang.NoSuchMethodError at runtime, on the very first line of the download thread.

The throw is caught by the thread's catch (error: Throwable) and reported as a generic update_failed carrying only the exception message, so from Dart it is indistinguishable from a network failure or a hash mismatch. Every update attempt on Android 5.0 to 6.0 fails this way, before a single byte is read.

packages/ndk_flutter/android/src/main/kotlin/relaystr/ndk/DartNdkPlugin.kt
```diff
@@ -42,14 +54,328 @@ class DartNdkPlugin : FlutterPlugin, MethodCallHandler, ActivityAware,
...
+                            "${_context.packageName}.ndk_updates",
+                            apk,
+                        )
+                        val intent = Intent(Intent.ACTION_INSTALL_PACKAGE, uri).apply {
+                            addFlags(Intent.FLAG_GRANT_READ_URI_PERMISSION)
+                            addFlags(Intent.FLAG_ACTIVITY_NEW_TASK)
+                            putExtra(Intent.EXTRA_NOT_UNKNOWN_SOURCE, true)
+                        }
+                        _context.startActivity(intent)
+                        completeUpdateOnMainThread(operation, "awaitingUserAction")
```
nogringo, line 277: Downloaded APKs are never pruned from the cache directory
completeUpdateOnMainThread deletes operation.apk only when the operation was cancelled or failed.

packages/ndk_flutter/lib/widgets/app_update/n_release_engagement.dart
```diff
@@ -0,0 +1,466 @@
...
+              .toList()
+            ..sort((a, b) => b.createdAt.compareTo(a.createdAt));
+      final authors = {
+        release.event.pubKey,
+        ...commentEvents.map((event) => event.pubKey),
+        ...satsByPubkey.keys,
+      };
+      final reactionCount = unique.where((event) => event.kind == 7).length;
+      _setState(
+        NReleaseEngagementState(
```
nogringo: The intermediate _setState in _load builds a fresh state instead of copyWith

_load sets state three times. The first and the last go through _state.copyWith, deliberately preserving everything they do not touch. The one in the middle, after the events are parsed and before metadata resolves, constructs a brand new NReleaseEngagementState, so every field it does not name falls back to its default.

Three fields are lost that way. publisher goes back to null, so the publisher row in the details sheet falls back to the npub and the avatar loses its picture until the metadata query returns.

nogringo (changes requested):

packages/ndk_flutter/lib/widgets/app_update/n_release_engagement.dart
nogringo: This can be solved with a synchronous metadata cache which ndk will need at some point.

packages/ndk_flutter/android/src/main/AndroidManifest.xml
```diff
@@ -7,4 +7,15 @@
             <data android:scheme="nostrsigner" />
         </intent>
     </queries>
+    <application>
+        <provider
+            android:name="androidx.core.content.FileProvider"
```
nogringo: Declaring androidx.core.content.FileProvider itself breaks any app that already has a FileProvider

packages/ndk/lib/domain_layer/usecases/software/software.dart
```diff
@@ -0,0 +1,297 @@
...
+        asset.event.id: asset,
+    };
+    return {
+      for (final release in releaseList)
+        release.event.id: release.assets
+            .map((reference) => assetsById[reference.eventId])
+            .whereType<SoftwareAsset>()
+            .where(
+              (asset) =>
+                  asset.identifier == release.identifier &&
```
nogringo: This asset filter contradicts NIP-82 and silently hides releases

packages/ndk_flutter/android/src/main/kotlin/relaystr/ndk/DartNdkPlugin.kt
```diff
@@ -42,14 +54,334 @@ class DartNdkPlugin : FlutterPlugin, MethodCallHandler, ActivityAware,
...
+                            if (operation.cancelled.get()) break
+                            val count = input.read(buffer)
+                            if (count < 0) break
+                            downloaded += count
+                            check(downloaded <= byteLimit) {
+                                "APK exceeds download size limit"
+                            }
+                            output.write(buffer, 0, count)
+                            digest.update(buffer, 0, count)
+                            if (total > 0) emitProgress(downloaded.toDouble() / total)
```
nogringo: Progress is emitted once per 8 KB read, and never at all without Content-Length

DEFAULT_BUFFER_SIZE is 8192, so this fires once per buffer. A 50 MB APK produces roughly 6400 events, and each one allocates a new Handler, posts a Runnable to the main looper, crosses the method channel, gets pushed into the broadcast stream in AndroidPackageInstaller, and reaches _setState in the controller, which calls notifyListeners and rebuilds every ListenableBuilder bound to it: banner, tile, version widget and sheet. All of this while the user is looking at a progress bar that cannot move by more than a pixel between two consecutive events.

Throttling on the native side would keep the same UX for a fraction of the cost: track the last emitted value, emit only when the rounded percentage changes or when a short interval has elapsed, and always emit the final value. The Handler can also be hoisted to a field instead of being allocated per chunk.

The other half of this line is the opposite problem. When the server does not send Content-Length, which is what a chunked response from a Blossom endpoint looks like, total stays at minus one and no progress is ever emitted. downloadAndInstall has already set progress to zero on the Dart side, and the sheet renders LinearProgressIndicator(value: state.progress), so the user gets a determinate bar frozen at empty for the whole download. That reads as a hang. Leaving progress null in that case would give an indeterminate bar, which is the honest rendering of an unknown total.

packages/ndk_flutter/lib/widgets/app_update/n_release_engagement.dart
```diff
@@ -0,0 +1,486 @@
...
+        zapTargetApp?.publisher,
+      }.whereType<String>().toList(growable: false);
+      final responses = await Future.wait([
+        ndk.requests
+            .query(
+              name: 'software-release-zaps-reactions',
+              explicitRelays: relays.isEmpty ? null : relays,
+              filter: Filter(
+                kinds: const [9735, 7],
+                pTags: publishers,
```
nogringo: The zaps and reactions query is scoped to the publisher, not to the releases

This asks every relay for all kind 9735 and kind 7 events p-tagging the publisher since the first release, capped at 1000, then _isRelevantEngagement throws away everything that does not reference a release or the application. The two comment queries right below are correctly scoped with #A and #a, so the asymmetry is visible in the same Future.wait.

packages/ndk_flutter/android/src/main/kotlin/relaystr/ndk/DartNdkPlugin.kt
```diff
@@ -42,14 +54,334 @@ class DartNdkPlugin : FlutterPlugin, MethodCallHandler, ActivityAware,
...
+                        check(expectedSize == null || downloaded == expectedSize) {
+                            "Downloaded APK size mismatch"
+                        }
+                    }
+                }
+                operation.input = null
+                val actualHash = digest.digest().joinToString("") { "%02x".format(it) }
+                check(actualHash == expectedHash) { "Downloaded APK SHA-256 mismatch" }
+
+                val flags = android.content.pm.PackageManager.GET_SIGNING_CERTIFICATES
```
nogringo: The API 28 branch landed in installedSoftware but not here

installedSoftware now picks GET_SIGNATURES below API 28, which is the fix for my earlier comment. This call site still passes GET_SIGNING_CERTIFICATES unconditionally, and the module declares minSdk = 21, so API 21 to 27 is still in scope.

Worth deciding at the same time whether Android 8.1 and older are meant to be supported at all here. If not, an explicit check with a clear message would beat a certificate mismatch after a full download.

nogringo (changes requested):

packages/ndk/lib/domain_layer/usecases/software/software.dart
```diff
@@ -0,0 +1,297 @@
...
+          name: 'software-app',
+        )
+        .future;
+    final parsed = events
+        .map(_parseApp)
+        .whereType<SoftwareApp>()
+        .where((candidate) => candidate.ref.identifier == app.identifier)
+        .toList();
+    parsed.sort((a, b) => b.event.createdAt.compareTo(a.event.createdAt));
+    return parsed.firstOrNull;
```
nogringo, line 44: getApp returns null for both "not published" and "no relay answered"

getReleases right below distinguishes the two: it inspects response.relayOutcomes and throws SoftwareDiscoveryException when nothing came back and no relay reached EOSE. getApp collapses both into null, so a caller cannot tell an application that was never published from a total relay failure.

packages/ndk/lib/domain_layer/usecases/software/software.dart
```diff
@@ -0,0 +1,299 @@
...
+        .map((asset) => asset.eventId)
+        .toSet();
+    if (ids.isEmpty) {
+      return {for (final release in releaseList) release.event.id: const []};
+    }
+    final hintedRelays = releaseList
+        .expand((release) => release.assets)
+        .map((asset) => asset.relayHint)
+        .whereType<String>();
+    final explicitRelays = {...?relays, ...hintedRelays};
```
nogringo: Relay selection here is delegated to the engine, but only one of the two engines does author-based (NIP-65 outbox) routing, so the behaviour differs depending on how NDK is wired.

Could the use case resolve the publisher's write relays itself rather than relying on the active engine?

packages/ndk/lib/domain_layer/usecases/software/software.dart
```diff
@@ -0,0 +1,299 @@
...
+    final events = await _requests
+        .query(
+          filter: Filter(ids: ids.toList(), kinds: const [softwareAssetKind]),
+          explicitRelays: explicitRelays.isEmpty ? null : explicitRelays,
```
nogringo: Relay hints should be used in addition to the resolved relay list, not instead
of it.

packages/ndk/lib/domain_layer/usecases/software/software.dart
```diff
@@ -0,0 +1,299 @@
...
+        final versionCode = asset.versionCode;
+        if (asset.mimeType != androidPackageMimeType ||
+            asset.url == null ||
+            asset.identifier != installed.packageId ||
+            versionCode == null ||
+            (asset.minPlatformVersion ?? 0) > installed.platformVersion ||
+            (asset.minAllowedVersionCode ?? 0) > installed.versionCode ||
+            (asset.variant != null && asset.variant != installed.variant) ||
+            !_supportsPlatform(asset.platforms, installed.platforms) ||
+            !_hasMatchingCertificate(
```
nogringo, line 239: Only the first `url` tag is kept, so there is no fallback if it fails. Assets usually publish more than one: on 60 kind 3063 events from relay.zapstore.dev, 50 carry two urls, GitHub first and the cdn.zapstore.dev Blossom mirror second. Keeping all of them and trying each in order would use that mirror, and the sha256 is verified afterwards anyway.

The publisher's kind 10063 server list could be appended as a last resort.

packages/ndk_flutter/lib/widgets/app_update/n_app_update_controller.dart
```diff
@@ -0,0 +1,594 @@
...
+        return;
+      }
+      final installed =
+          _state.installed ?? await installer.getInstalledSoftware();
+      if (_disposed) return;
+      final previousUpdate = _state.update;
+      if (previousUpdate != null &&
+          installed.packageId == previousUpdate.asset.identifier &&
+          installed.versionCode >= previousUpdate.asset.versionCode!) {
+        if (!_canCommitEvaluation(generation)) return;
```
nogringo, line 323: The generation guard runs before the `await` on line 306, so a concurrent evaluation that commits during that await gets overwritten by this stale one. The main path re-checks after its own await (lines 356-358), this branch does not.

nogringo (changes requested):

packages/ndk/lib/domain_layer/usecases/software/software.dart
frnandu: Since we plan to merge both engines and the default will use outbox I suggest we don't add duplicate write relays discover in every possible usecase.
Wdyt @1-leo ?

packages/ndk_flutter/android/src/main/kotlin/relaystr/ndk/DartNdkPlugin.kt
```diff
@@ -42,14 +55,376 @@ class DartNdkPlugin : FlutterPlugin, MethodCallHandler, ActivityAware,
...
+                    "APK certificate is incompatible with installed app"
+                }
+
+                _mainHandler.post {
+                    if (operation.cancelled.get()) {
+                        completeUpdateOnMainThread(operation, "cancelled")
+                        return@post
+                    }
+                    try {
+                        val uri = FileProvider.getUriForFile(
```
nogringo, line 265: The plugin declares `minSdk = 21`, but this hands the installer a `content://` URI from `FileProvider`. The system package installer only accepts `content://` from Android 7 (API 24). On Android 5 and 6 it only takes `file://`, so the install screen fails with a parse error.

Simplest fix: treat in-app install as unsupported when `Build.VERSION.SDK_INT < Build.VERSION_CODES.N`, and hide the update action on the Dart side (`platformVersion` is already sent back by `installedSoftware`).

packages/ndk/lib/domain_layer/usecases/software/software.dart
```diff
@@ -0,0 +1,312 @@
...
+
+  bool _supportsPlatform(List<String> asset, List<String> installed) =>
+      asset.isEmpty || asset.any(installed.contains);
+
+  bool _hasMatchingCertificate(List<String> asset, List<String> installed) {
+    final normalized = installed.map((value) => value.toLowerCase()).toSet();
+    return asset.any((value) => normalized.contains(value.toLowerCase()));
+  }
+
+  List<SoftwareRelease> _validReleases(
```
nogringo, line 262: Release events are addressable (`kind:30063`, `d` = `i@version`), so relays can hold different copies of the same release. `Requests.query` only drops duplicate event ids, so two copies with different ids both get through, and `_validReleases` keeps both.

`watchReleases` handles this already: it keeps only the newest copy per version. `getReleases` doesn't, so the two can return different lists for the same app.

packages/ndk_flutter/lib/widgets/app_update/n_app_update_controller.dart
```diff
@@ -0,0 +1,595 @@
...
+    if (_preservesInstallerState && !allowDuringInstaller) {
+      return;
+    }
+    final completer = Completer<void>();
+    _checkCompleter = completer;
+    final totalTimer = Stopwatch()..start();
+    _checking = true;
+    _setState(
+      _state.copyWith(
+        status: NAppUpdateStatus.checking,
```
nogringo, line 238: Resuming the app during a download breaks the download UI. `didChangeAppLifecycleState` calls `checkNow(allowDuringInstaller: true)`, which skips the `_preservesInstallerState` guard, sets the status to `checking` and clears the progress. `_evaluate` then returns right away because `_installOperationActive` is still true, so nothing puts `downloading` back.

What the user sees in `NAppUpdateSheet` / `NAppUpdateDialog` while the download keeps running:
- `busy` is false, so the progress bar disappears and Cancel turns into Later, which calls `dismiss()` and closes the dialog.
- The install button is enabled again, but `downloadAndInstall` returns early.
- Progress is only emitted when the server sends `Content-Length`. Without it, the UI stays like this until the download finishes.
- If the relay query fails, `_fail` shows an error for a download that is still in progress.

Since `_evaluate` throws away the result during a download anyway, the resume check could just return early when `_installOperationActive` is true. Returning from the system installer still works, because the flag is already false by then.

nogringo (changes requested):

packages/ndk/lib/domain_layer/usecases/software/software.dart
1-leo: agree

PR #804, opened 2026-09-10
feat(cache): add loadHiddenEvents

nogringo:
An app had no way to reach a superseded version of a replaceable event, or one its author deleted. Both stay in the cache, but loadEvents hides them and loadEvent only answers to a caller who already holds the id, which is exactly what it lacks. loadHiddenEvents queries them by filter, including by conflict domain, and pairs each event with the cache state that says why it is hidden.

Two deletion gaps close on the way. The objectbox backend ignored coordinate deletions, its check only looked at `e` tags. And a coordinate for a non addressable replaceable kind now matches whether its `a` tag carries a trailing empty d-tag segment or not.

BREAKING CHANGE: CacheManager gains loadHiddenEvents, so a third-party cache backend has to implement it.

## Summary by CodeRabbit

* **New Features**
  * Added support for retrieving cached events hidden because they were superseded, deleted, or expired.
  * Added filters for hidden-event reasons and event coordinates.
  * Added metadata explaining why an event is hidden, including deletion details.
* **Bug Fixes**
  * Improved deletion matching for replaceable and addressable events.
  * Applied result limits after visibility filtering.
* **Documentation**
  * Documented hidden-event retrieval and cache behavior when displaying history.

### DIFF
--- a/packages/drift/lib/src/drift_cache_manager.dart
+++ b/packages/drift/lib/src/drift_cache_manager.dart
@@ -18,8 +18,8 @@ import 'package:ndk/domain_layer/entities/wallet/wallet_transaction.dart';
 import 'package:ndk/domain_layer/entities/wallet/wallet_type.dart';
 import 'package:ndk/domain_layer/repositories/wallets_repo.dart';
 import 'package:ndk/ndk.dart';
-import 'package:ndk/shared/nips/nip01/event_kind_classification.dart';
 import 'package:ndk/shared/nips/nip01/event_eviction_planner.dart';
+import 'package:ndk/shared/nips/nip01/event_visibility_resolver.dart';
 
 import 'database/database.dart';
 
@@ -36,6 +36,10 @@ class DriftCacheManager extends WalletsRepo implements CacheManager {
   final NdkCacheDatabase _db;
   String? _defaultWalletIdForReceiving;
   String? _defaultWalletIdForSending;
+
+  late final EventVisibilityResolver _visibility = EventVisibilityResolver(
+    _loadRawEvents,
+  );
 
   DriftCacheManager(this._db) {
     unawaited(_initializeWalletDefaults());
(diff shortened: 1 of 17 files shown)


packages/ndk/lib/domain_layer/entities/event_cache_records.dart
```diff
@@ -558,14 +588,26 @@ class EventCacheStateRecord {
...
     for (final event in deletionEvents) {
       if (event.pubKey != target.pubKey) continue;
       if (event.getTags('e').contains(target.id.toLowerCase())) {
         return event;
       }
-      if (replaceableConflictKey != null &&
+      if (coordinates.isNotEmpty &&
           event.createdAt >= target.createdAt &&
-          event.getTags('a').contains(replaceableConflictKey)) {
+          event.getTags('a').any(coordinates.contains)) {
```
frnandu, line 610: 🔴 bug: getTags("a") lowercases the entire coordinate, while the target conflict key preserves the case-sensitive d-tag. A deletion for d-tag Article-A becomes article-a and never matches, leaving author-deleted content visible. Compare the raw a-tag value and normalize only kind/pubkey; add a mixed-case d-tag regression test.

packages/drift/lib/src/drift_cache_manager.dart
```diff
@@ -142,6 +146,63 @@ class DriftCacheManager extends WalletsRepo implements CacheManager {
...
     int? limit,
+  }) async {
+    var events = await _loadRawEvents(
+      ids: ids,
+      pubKeys: pubKeys,
+      kinds: kinds,
+      tags: tags,
+      since: since,
+      until: until,
+      search: search,
```
frnandu, line 157: 🔴 risk: omitting limit here makes loadEvents(limit: 1) materialize every matching row before filtering, sorting, and running another context query. ObjectBox and Sembast have the same regression, so routine bounded reads can become O(cache size) in time and memory. Use paged overfetch until enough visible rows are found, or make visibility queryable without removing the backend bound.

frnandu (changes requested):
Requesting changes for two blocking issues: case-sensitive deletion coordinates and unbounded limited reads.

1-leo (approved):
api looks good

packages/drift/lib/src/drift_cache_manager.dart
nogringo: Checked this against the backends and I don't think the regression is there.

Drift and ObjectBox never bounded this read: on master both already materialized every matching row and applied the limit in memory, so their cost is unchanged by this PR.

Sembast did pass the limit to the Finder, but that bound was inert. Sembast only applies a limit during the scan when the Finder has no sort order (store_impl.dart, _FinderData.usePreordered). We always pass SortOrder('created_at', false), so it walks every record in the store, sorts all matches, then trims. Dropping the pushdown costs the deserialization of the extra matches, out of a store sembast already holds fully in memory. It buys back the correctness bug this PR fixes, where the limit was spent on events the visibility rules then removed.

The candidate set is also always scoped: CacheRead only resolves filters that carry authors or ids, so a broad unbounded read never reaches the cache.

The part I do think is worth changing is the context read: above 100 authors it drops the pubkey constraint entirely and scans every event of the candidate kinds. Batching the authors would keep the bound at any size.

packages/ndk/lib/domain_layer/entities/event_cache_records.dart
```diff
@@ -458,6 +458,58 @@ class EventCacheStateRecord {
...
+  /// names a kind that has no conflict domain.
+  ///
+  /// Only the pubkey is case folded: NIP-01 mandates lowercase hex there, while
+  /// a d-tag is an arbitrary string relays match byte for byte.
+  static String? normalizeConflictKey(String coordinate) {
+    final parts = coordinate.split(':');
+    if (parts.length < 2) return null;
+    final kind = int.tryParse(parts[0]);
+    if (kind == null || parts[1].isEmpty) return null;
+    final pubKey = parts[1].toLowerCase();
```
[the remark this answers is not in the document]
nogringo: Went the other way on this one, and dropped the fold instead.

The asymmetry is real, but the fix as proposed only covers the conflict key. An event stored with a non-lowercase pubkey stays unreachable through an authors filter anyway, since the backends compare the stored column raw, so lowercasing in the builder would not make that event matchable. It would just widen what we accept.

NIP-01 is explicit that the input is invalid: pubkey is "32-bytes lowercase hex-encoded", and filter lists "MUST contain exact 64-character lowercase hex values". Nothing in ndk can produce another spelling either, nip19 decoding goes through hex.encode, which is lowercase.

frnandu (approved):

frnandu (changes requested):
some metadata test is failing, idk why

frnandu (approved):

PR #805, opened 2026-09-11
feat: let a broadcast say which identity it may be attributed to

### LINKED ISSUE #665: authenticateAs for broadcast

nogringo:
The auth parameter now reaches the write path, so a caller can refuse to be attributable, or bind the connection before the relay asks. Only the canonical policy is persisted for background retries: a delivery whose identity no account can sign for waits in needsAction rather than going out as the logged account.

Closes #665.

## Summary by CodeRabbit

* **New Features**
  * Added NIP-42 relay authentication options for event broadcasts, including anonymous, allowed-identity, and required-identity policies.
  * Broadcasts now apply authentication consistently across direct delivery, retries, reactions, and deletions.
  * Added handling for required identities that cannot sign, with resumable pending deliveries.

* **Bug Fixes**
  * Preserved authentication settings for pending broadcasts across application restarts.

* **Documentation**
  * Documented relay authentication behavior, policies, errors, and restart handling.

### DIFF
--- a/packages/ndk/lib/domain_layer/usecases/broadcast/pending_broadcast_delivery.dart
+++ b/packages/ndk/lib/domain_layer/usecases/broadcast/pending_broadcast_delivery.dart
@@ -251,9 +251,20 @@ class PendingBroadcastDelivery {
       if (existingTarget != null) {
         // only the relays of this call are re-attributed; the same event may
         // be on its way to another relay as somebody else
-        return auth == null
-            ? existingTarget
-            : existingTarget.copyWith(authCanonical: auth.canonical);
+        if (auth == null) {
+          return existingTarget;
+        }
+        final reattributed = existingTarget.copyWith(
+          authCanonical: auth.canonical,
+        );
+        // naming an identity is what a target parked for a missing one was
+        // waiting for, so it can move again
+        return _isParkedForIdentity(reattributed)
+            ? reattributed.copyWith(
+                state: RelayDeliveryState.pending,
+                lastError: null,
+              )
+            : reattributed;
       }
       return RelayDeliveryTarget(
         eventId: event.id,
@@ -417,6 +428,10 @@ class PendingBroadcastDelivery {
 
         if (!onlyDue) {
           return true;
+        }
+
+        if (_isParkedForIdentity(target)) {
+          return false;
         }
 
         return target.nextRetryAt == null || target.nextRetryAt! <= now;
@@ -643,6 +658,15 @@ class PendingBroadcastDelivery {
     await _cacheManager.removeEventDeliveryRecord(eventId);
   }
 
+  /// A target waiting for an identity the app has to name again. Nothing about
+  /// it changes on its own, so it must not make its relay due: it would
+  /// reconnect and rewrite the same state on every retry interval, forever.
+  /// A normal auth-required refusal carries a [RelayDeliveryTarget.nextRetryAt]
+  /// and stays retryable.
+  bool _isParkedForIdentity(RelayDeliveryTarget target) =>
+      target.state == RelayDeliveryState.authRequired &&
+      target.nextRetryAt == null;
+
   Future<Set<String>> _relayUrlsWithDuePendingTargets() async {
     final now = Nip01Event.secondsSinceEpoch();
     final targets = await _cacheManager.loadRelayDeliveryTargets(
@@ -652,6 +676,9 @@ class PendingBroadcastDelivery {
     final relayUrls = <String>{};
     for (final target in targets) {
       if (target.state == RelayDeliveryState.permanentFailure) {
+        continue;
+      }
+      if (_isParkedForIdentity(target)) {
         continue;
       }
       if (target.nextRetryAt != null && target.nextRetryAt! > now) {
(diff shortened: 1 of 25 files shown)


packages/ndk/lib/domain_layer/usecases/jit_engine/relay_jit_broadcast_strategies/relay_jit_broadcast_specific.dart
```diff
@@ -39,48 +38,23 @@ class RelayJitBroadcastSpecificRelaysStrategy {
...
-            cacheManager: cacheManager,
-            event: eventToPublish,
-          )) {
+        if (!relayManager.isRelayConnected(relayUrl)) {
+          final success = (await relayManager.connectRelay(
+            dirtyUrl: relayUrl,
+            connectionSource: ConnectionSource.broadcastSpecific,
+            connectTimeout: 1,
+          ))
+              .first;
```
frnandu, line 47: 🔴 bug: `RelayAuthRequire` opens this anonymous socket before opening its bound socket, so a relay connection limit can reject the only socket allowed to carry the event. Remove anonymous preconnection here and in own/inbox strategies; make `connectionForBroadcast` open or reuse only policy-selected connection.

packages/ndk/lib/domain_layer/usecases/broadcast/pending_broadcast_delivery.dart
```diff
@@ -226,6 +237,7 @@ class PendingBroadcastDelivery {
         lastSignAttemptAt: existing?.lastSignAttemptAt,
         nextSignRetryAt: existing?.nextSignRetryAt,
         lastSignError: existing?.lastSignError,
+        authCanonical: auth?.canonical ?? existing?.authCanonical,
```
frnandu, line 240: 🟡 risk: rebroadcasting same pending event with `auth: null` preserves previous `authCanonical`, while live map also retains old policy; later retries can authenticate as stale identity. Clear live entry and persist null when current broadcast omits auth.

packages/ndk/lib/domain_layer/usecases/broadcast/pending_broadcast_delivery.dart
```diff
@@ -471,14 +486,87 @@ class PendingBroadcastDelivery {
...
+  ) async {
+    Logger.log.w(
+      () => 'delivery ${record.eventId} to ${target.relayUrl} needs '
+          '${record.authCanonical}, which no account can sign for',
+    );
+    await _cacheManager.saveRelayDeliveryTargets([
+      target.copyWith(
+        state: RelayDeliveryState.authRequired,
+        nextRetryAt: null,
+        lastError: 'No available account for ${record.authCanonical}',
```
frnandu, line 564: 🟡 risk: parked `authRequired` target gets `nextRetryAt: null`, but both due predicates treat null as immediately due, causing reconnect and storage writes every retry interval. Exclude parked auth targets until caller supplies identity.

packages/ndk/lib/domain_layer/usecases/broadcast/pending_broadcast_delivery.dart
```diff
@@ -195,10 +202,14 @@ class PendingBroadcastDelivery {
     required Nip01Event event,
     required Iterable<String> relayUrls,
     required bool requiresInteractiveSigning,
+    RelayAuth? auth,
   }) async {
     if (_stopped) {
       return;
     }
+    if (auth != null) {
+      _authPolicies[event.id] = auth;
```
frnandu, line 211: 🟡 risk: every authenticated event remains in `_authPolicies` after delivery, failure, or discard, retaining `RelayAuth`, `Account`, and signer references for process lifetime. Evict policy once all targets are terminal and from `_discardEventDelivery`.

packages/ndk/test/mocks/mock_relay.dart
```diff
@@ -65,6 +65,29 @@ class MockRelay {
...
 
+  /// every EVENT received, with the socket that carried it
+  final List<_ReceivedEvent> _receivedEventConnections = [];
+
+  /// Whether [socket] is authenticated as [pubkey] *now*, not when the event
+  /// arrived. A connection bound to an identity sends before the relay has
+  /// accepted its AUTH, so judging at arrival time would call a bound
+  /// connection anonymous.
+  bool _socketCarries(WebSocket socket, String pubkey) =>
+      _authenticatedPubkeys[socket]?.contains(pubkey) ?? false;
```
frnandu, line 76: 🟡 test gap: attribution lookup reads socket map later; socket cleanup can remove entry before assertion and classify authenticated event as anonymous. Store connection-owned pubkey set with received event, preserving later AUTH mutation and post-close assertions.

frnandu (changes requested):
Requesting changes for authenticated-connection routing and pending-delivery lifecycle issues. Required-auth broadcasts can fail on relays with connection limits; remaining comments cover stale policy reuse, repeated parked retries, retained signer references, and a timing-dependent test helper.

packages/ndk/lib/domain_layer/usecases/broadcast/pending_broadcast_delivery.dart
nogringo: The stale-identity part was real and is fixed in 1ec9a444: the policy now lives on the relay target instead of on the event. `_authPolicies` is keyed by `RelayDeliveryTarget.key` (`eventId|relayUrl`), so broadcasting the same event to another relay under another identity no longer re-attributes the first one.

frnandu (approved):

1-leo (approved):

PR #818, opened 2026-09-12
feat: deterministic cashu quote key

### LINKED ISSUE #809: Cashu: mint quote lock key is random and unrecoverable; paid quotes are abandoned when the app closes
## Summary

The bundled NDK Cashu wallet can permanently strand user funds at a mint. A paid
mint quote is abandoned if the app is not open when the Lightning payment lands,
and the key needed to redeem that quote is random and stored only next to the
local quote record — so once local data is lost, the sats are unrecoverable by
anyone, including the mint.

This is not hypothetical. Two users of an SK coordinator hit it in ten days:
263 714 sats (now unrecoverable) and 147 542 sats (possibly still recoverable).
In both cases the coordinator's Lightning payment to `mint.minibits.cash`
SUCCEEDED with a preimage, and the mint reports the quote as `paid / 0 issued`.

The Minibits operator asked us directly to report this upstream, and phrased the
two requirements themselves:

> - wallet MUST derive mint quote keys from the seed if it owns them (ie user
>   does not provide their own)
> - wallet MUST handle intermittent network and other failures to mint and keep
>   request payloads, then retry the mint requests using mint's idempotent cache
>   or otherwise provide user the option to retry the minting.

Versions: `ndk 0.8.4-dev.8` (bitblik app depends on `ndk: ^0.8.4-dev.0`).

## Defect 1 — the quote lock key is random and can never be re-derived

- `packages/ndk/lib/data_layer/repositories/cashu/cashu_repo_impl.dart:164`
  `getMintQuote()` creates `CashuKeypair.generateCashuKeyPair()` for every quote
  and sends `quoteKey.publicKey` to the mint as the NUT-20 `pubkey` lock.
- `packages/ndk/lib/domain_layer/usecases/cashu/cashu_keypair.dart:18`
  the private key is `Helpers.getSecureRandomHex(32)` — pure entropy, unrelated
  to the wallet seed.
- `packages/ndk/lib/domain_layer/entities/cashu/cashu_quote.dart:66`
  the keypair is persisted only inside the quote record's JSON.

Because the mint locks issuance to that pubkey, only the holder of that private
key can ever mint the ecash. The seed does not reproduce it — seed restore
returns already-minted proofs, not this key. So if the local quote record is lost
(cleared browser storage, reinstall, restored DB), the paid quote becomes
permanently unredeemable. The mint cannot help either: NUT-20 requires the
signature it no longer has a counterpart for.

This is what makes the failure *terminal* rather than merely annoying.

## Defect 2 — minting is not resumable, which is what strands the quote

- The whole flow lives in one async generator in
  `packages/ndk/lib/domain_layer/usecases/cashu/cashu.dart`: it polls for payment
  on `CashuConfig.FUNDING_CHECK_INTERVAL` (~line 898) and only afterwards calls
  `mintTokens()` (~line 911).
- `_pendingTransactions` (line 92) is an in-memory `Set`. There is no startup
  scan, and nothing ever revisits a persisted pending quote.
- Consequently, if the app is closed or the page reloaded while the payment is in
  flight, the stream dies, the payment arrives, and no code path ever requests
  the signatures. The UI offers no retry.

Defect 2 creates the stranded quote; Defect 1 makes it unrecoverable.

## Why this matters more than a normal bug

The amounts are not zap-sized. Users reach this state by doing nothing wrong —
closing an app during a payment is ordinary behaviour. From the user's side it
looks exactly like theft by the coordinator or the mint, and the coordinator has
to prove with preimages that the money did leave and did arrive. The mint
operator's own words: "providing half-baked wallet is much worse than providing
none."

## Suggested fix

1. **Derive the quote key from the seed.** If the wallet owns the key, it must be
   reconstructible: a dedicated derivation path indexed by a quote counter, so
   any restored wallet can re-sign for any quote it ever created. A key supplied
   by the user stays the user's responsibility.
2. **Persist pending mint quotes and resume them.** Re-drive them on startup and
   on wallet open, and expose an explicit "retry minting" action. The mint's mint
   request is idempotent/cached, so re-requesting is safe.
3. **Until 1 ships**, warn before anything that clears local data while a quote is
   pending, and surface pending quotes in the UI so they are not invisible.

Fix 1 alone would have saved both users: even with no retry button, a
seed-derived key lets the ecash be minted later, by hand if necessary.

## Proposed patch

I have implemented fix 1 and the recovery path on a branch and can open a PR:

- `DerivationType.quoteKey(2)` added to the existing HMAC-SHA256 KDF, so quote
  keys reuse the scheme already used for NUT-13 secrets and blinding factors,
  with a distinct type byte. Bound to `sha256(mintUrl)` rather than a keyset id,
  because no keyset is chosen yet when the quote is created.
- `CashuKeyDerivation.deriveQuoteKey({seedBytes, counter, mintUrl})` and
  `CashuKeypair.fromPrivateKeyHex()`.
- `getMintQuote()` no longer generates entropy; it takes the derived keypair and
  the counter it came from. The counter is obtained from the existing
  `getAndIncrementDerivationCounter` under a reserved slot `quote-key`, which
  cannot collide with a real keyset id since those are always hex.
- `CashuQuote.quoteKeyCounter` is persisted; quotes written before this land
  deserialize as `-1`, so the change is backward compatible.
- `Cashu.recoverQuoteKey({mintUrl, lockedPubkey, maxScan})` walks counters until
  the derived pubkey matches the one the mint locked the quote to. This is what
  turns "local data gone" from fatal into recoverable.

Happy to adjust the derivation message format if you would rather bind it
differently — the important property is only that it is reproducible from the
seed.

## Evidence

Available on request (trade ids, quote ids, payment hashes, preimages, and the
mint's own `paid / 0 issued` export). Withheld here only because this is a public
tracker and the data identifies individual trades.

1-leo:
Implementation for seed derived cashu quote keys so funds can be recovered by seed only

## Summary by CodeRabbit

- **New Features**
  - Added deterministic Cashu mint-quote key derivation and recovery from the wallet seed.
  - Pending quotes now refresh automatically and can recover missing quote keys.
  - Paid-but-incomplete quotes can be recovered and completed during restoration.
  - Added progress reporting during quote recovery and completion.
  - Quote-key counters are preserved across wallet export and import.
  - Added recovery for mint quotes by quote ID, including quotes missing from local wallet state.

- **Updates**
  - Mint-quote requests now require the associated quote key and derivation counter.


bit-blik (dismissed):

packages/ndk/lib/domain_layer/usecases/cashu/cashu.dart
```diff
@@ -623,6 +655,432 @@ class Cashu {
...
+    } else {
+      draft = existing.copyWith(
+        state: WalletTransactionState.pending,
+        qoute: quote,
+      );
+    }
+
+    // complete the mint: waits for payment, marks the transaction `completed`
+    // (or `failed`) and saves the minted proofs
+    final events = await retrieveFunds(draftTransaction: draft).toList();
```
frnandu, line 1080: [P1] Restore proof counters before minting a recovered quote

After local state loss, `recoverAndCompleteQuote()` calls `retrieveFunds()` with a fresh cache, so proof derivation starts at counter zero. If this seed has previously minted against the selected keyset, the recovery request reuses previously signed blinded outputs and the mint rejects it with error 11003 (Outputs already signed). I reproduced the recovery request generating the same blinded output as the original counter-zero derivation. Restore or otherwise safely advance the proof counters before minting the recovered quote.

packages/ndk/lib/domain_layer/usecases/cashu/cashu.dart
```diff
@@ -623,6 +655,432 @@ class Cashu {
...
+          );
+          continue;
+        }
+
+        final keypair = await _cashuKeyDerivation.deriveQuoteKey(
+          seedBytes: Uint8List.fromList(_cashuSeed.getSeedBytes()),
+          counter: counter,
+        );
+
+        if (keypair.privateKey != quote.quoteKey.privateKey ||
```
frnandu, line 882: [P1] Preserve valid quote keys when the configured seed differs

A stored derivation counter does not establish that the currently configured seed created this quote. When pending transactions survive a seed change, startup derives a different key at the same counter and this branch overwrites the valid stored private and public keys. I reproduced startup replacing a valid quote key with a key from another seed, preventing redemption with the saved record. Preserve valid stored keys and verify a recovered public key against the quote lock before saving a replacement.

packages/ndk/lib/domain_layer/usecases/cashu/cashu.dart
```diff
@@ -623,6 +655,432 @@ class Cashu {
...
+        walletType: WalletType.CASHU,
+        state: WalletTransactionState.pending,
+        mintUrl: mintUrl,
+        method: method,
+        usedKeysets: [keyset],
+        qoute: quote,
+        initiatedDate: DateTime.now().millisecondsSinceEpoch ~/ 1000,
+      );
+    } else {
+      draft = existing.copyWith(
```
frnandu, line 1072: [P2] Fill missing minting metadata on existing recovery records

The existing-record branch updates only the quote and transaction state. If a surviving quote record lacks `usedKeysets`, recovery throws `Used keysets is not specified in the transaction`; a missing `method` fails similarly even though this method has the payment method available. I reproduced the missing-keyset failure, while recovery without the existing record proceeds through the branch that fetches the required keyset. Populate missing method and active keyset metadata before calling `retrieveFunds()`.

frnandu (changes requested):
Requesting changes for three recovery issues, each reproduced separately. The 42 focused tests passed; live-mint integration tests were excluded.

packages/ndk/lib/domain_layer/usecases/cashu/cashu.dart
1-leo: yes that can happen when running `recoverAndCompleteQuote()` before `restore` on a seedphrase with empty local state.

I could do something like

``` dart
  Future<void> _advanceProofCounters({
    required String mintUrl,
    required String unit,
  }) async {
    final allKeysets = await _cashuKeysets.getKeysetsFromMint(mintUrl);
    final keysets = allKeysets.where((keyset) => keyset.unit == unit).toList();
    if (keysets.isEmpty) {
      return;
    }

    final cashuRestore = CashuRestore(
      cashuRepo: _cashuRepo,
      cashuKeyDerivation: _cashuKeyDerivation,
      cacheManager: _cacheManagerCashu,
      cashuSeed: _cashuSeed,
    );

    await for (final _ in cashuRestore.restoreAllKeysets(
      mintUrl: mintUrl,
      keysets: keysets,
    )) {}
  }
  ```
  
  and run this in `recoverAndCompleteQuote()` (effectively restoring everything on a quote restore)
  
However I think this should be up to the app developer, currently I dont see a scenario where you dont want to recover already minted tokens.
  
If run in sequence `restore() then recoverAndCompleteQuote()` its fine

File: packages/ndk/lib/domain_layer/usecases/cashu/cashu.dart (hunk 2 above)
Comment (1-leo) [reply packages/ndk/lib/domain_layer/usecases/cashu/cashu.dart]: a seed change is with the same wallet is not supported

File: packages/ndk/lib/domain_layer/usecases/cashu/cashu.dart (hunk 1 above)
Comment (frnandu) [reply packages/ndk/lib/domain_layer/usecases/cashu/cashu.dart]: Calling `restore()` first can be an explicit prerequisite for manual recovery, but it does not cover the automatic startup path. `_scheduleStartupResume()` calls `_completePaidPendingQuotes()` without restoring the proof counters, and the caller cannot await a restore before this constructor-triggered work begins. On f7d60994 I reproduced startup with the same seed, surviving pending transactions, and an empty proof cache generating the original counter-zero blinded output without any restore request. Please ensure the counters are safe before automatic minting, or gate automatic completion until restoration has finished. The manual recovery prerequisite should also be documented and enforced.

File: packages/ndk/lib/domain_layer/usecases/cashu/cashu.dart (hunk 2 above)
Comment (frnandu) [reply packages/ndk/lib/domain_layer/usecases/cashu/cashu.dart]: Keeping seed changes outside the supported scope is reasonable. The remaining concern is destructive handling of that unsupported state: startup currently replaces a valid stored quote key with a key derived from the mismatched seed. I reproduced that again on f7d60994. This does not require supporting seed migration: preserve a valid stored key, and reject a recovered key whose public key does not match the quote lock before saving it. An unsupported seed/state combination should fail without destroying the existing signing key.

File: packages/ndk/lib/domain_layer/usecases/cashu/cashu.dart (hunk 3 above)
Comment (frnandu) [reply packages/ndk/lib/domain_layer/usecases/cashu/cashu.dart]: Confirmed fixed in f7d60994: the existing-record branch now backfills the method and null/empty `usedKeysets`. My targeted reproduction with a matching existing quote record now reaches minting instead of failing on missing keysets. The checked-in existing-record test still has its `qoute` fixture commented out, so it does not exercise that branch; please restore the fixture so the regression is covered.

File: packages/ndk/lib/domain_layer/usecases/cashu/cashu.dart [hunk 4]
```diff
@@ -623,6 +657,485 @@ class Cashu {
...
+        continue;
+      }
+      try {
+        final freshState = await _cashuRepo.checkMintQuoteState(
+          mintUrl: tx.mintUrl,
+          quoteID: quote.quoteId,
+          method: method,
+        );
+
+        final refreshedTx = tx.copyWith(
```
frnandu: [P2] Prevent startup refresh from overwriting completed recovery

`tx` is a snapshot taken before the mint request. If explicit recovery finishes while that request is in flight, this save overwrites the newly completed transaction with the older pending snapshot. I reproduced this on f7d60994 by delaying the startup quote-state response, completing `recoverAndCompleteQuote()` against a surviving record with missing keysets, then releasing the response: the saved state changed from SUCCESS to PENDING and the recovered keysets became null again. Serialize refresh/recovery operations per quote and re-read or conditionally update the current record before saving, so a stale refresh cannot undo a completed recovery.

packages/ndk/lib/domain_layer/usecases/cashu/cashu.dart
```diff
@@ -623,6 +657,485 @@ class Cashu {
...
+
+    Logger.log.i(
+      () => 'Recovering quote keys for ${pendingFunding.length} pending '
+          'funding transactions on $mintUrl',
+    );
+
+    for (final tx in pendingFunding) {
+      final quote = tx.qoute!;
+      try {
+        final counter = quote.quoteKeyCounter >= 0
```
frnandu: [P2] Skip automatic counter scans for valid stored legacy keys

Every legacy pending quote has `quoteKeyCounter == -1`, so startup scans at least 1,000 derivations even when the quote already has its valid persisted private key. These derivations run on the calling isolate; awaiting their immediately completed futures does not yield to UI/timer events. On f7d60994, one legacy quote delayed a 10 ms timer until 6,447 ms, and the scan is repeated for each pending legacy quote. Preserve and use valid stored keys without scanning. Run genuinely necessary recovery scans in another isolate or cooperatively yield between batches so upgrading an existing wallet does not freeze its UI.

frnandu (changes requested):
Follow-up review of f7d60994: missing minting metadata is fixed. The proof-counter and stored-key concerns remain open as discussed in their threads. Two additional startup regressions are detailed inline: stale refreshes can undo completed recovery, and unnecessary legacy-key scans block the calling isolate. Validation: 43 focused tests passed; both new issues were reproduced separately with controlled local fixtures. Live-mint integration tests were excluded locally.

martinapoltergan:
Read through this — thank you for turning it around so fast. Notes from the
operator side, where the original incident happened.

**The derivation reads correct to me.** The counter slot is global
(`kQuoteKeyDerivationCounterSlot` passed as both `keysetId` and `mintUrl`), so
with `m/129373'/20'/0'/0'/<counter>` every quote gets a distinct key regardless
of mint, and the `129373'` purpose keeps it clear of NUT-13's `129372'`. I had
bound my own draft to `sha256(mintUrl)` with per-mint counters; with a global
counter that is redundant, and yours is the simpler design. Caching the hardened
parent per seed is the right call too — counter scans are the hot path.

**One gap I could not find covered: the counter after a seed-only restore.**

`setCashuSecretCounter` is carried through export/import, which is good. But a
user who restores from the seed phrase alone — no export file, which is exactly
the state both of our affected users were in — has no counter slot, so
`getAndIncrementDerivationCounter` starts at 0 and new quotes get keys that
older quotes already used.

To be clear about the blast radius, because it is smaller than it first looks:
funds are not at risk. NUT-20 signs over the specific quote, so there is no
cross-redemption, and recovery still works since the same key opens both. What
you get is (a) the mint seeing one lock pubkey across two quotes, which links
them, and (b) `_findQuoteKeyCounter` matching ambiguously.

The awkward part is that it cannot be fixed by scanning. NUT-13 can walk forward
because the mint will answer for derived blinded messages; NUT-20 has no "do you
hold a quote locked to this pubkey" lookup, so after a seed-only restore the
wallet has no way to learn how far the counter got. **That makes the counter
itself backup-critical state, not a local optimisation** — which is a concrete
argument for your nostr backup idea, and I would put the quote-key counter in it
explicitly rather than letting it ride along.

**On "the quoteId must still be retrieved from the mint operator if all state is
lost" — field data on how bad that is.** In the unrecoverable case behind #809,
the user could only obtain the quote id and the locked pubkey from a CSV export
the mint operator sent him by email, and only after several days of back and
forth. He had already proved control of the trading identity with a signed Nostr
event. So an API taking a `quoteId` is necessary, but in practice the user
usually cannot supply one: what actually unblocks them is either the mint
supporting lookup by the locked pubkey, or the wallet having backed up its quote
ids somewhere that survives the device. Your backup plan covers the second, and
is the only one of the two that does not need every mint to cooperate.

Worth noting the other half of the same surface: the only backup path today is
"copy to clipboard", which silently truncates on Android while reporting success
(#811). The user in the recovered case lost most of his backup that way and got
at it only by plugging the phone into a monitor to use desktop-mode clipboard. A
backup that is not a clipboard string fixes that too.

**Re the PR you offered to compare against:** mine is ~150 lines and only covers
defect 1 plus a `recoverQuoteKey()` scan; #818 supersedes it in every respect, so
I am not going to open it as noise. If you still want it as a cross-check the
branch is `fix/derivable-cashu-quote-key` and I am happy to push it — say the
word. Otherwise I would rather spend the effort testing #818 against the live SK
coordinator once it is ready, including the paid-but-unminted state that started
all this.

1-leo:
@martinapoltergan  The restoring of the quote counter after a wallet seed restore will not be part of this PR. Thx for highlighting the quote pubkey lookup ambiguity!
Right now the only (reasonable) solution for this would be a nostr backup. The issue of losing funds with pending quotes is mitigated by this PR and we want to get it out soon.
The nostr enabled backup will be a seperate PR, created an issue for it (https://github.com/relaystr/ndk/issues/829)

packages/ndk/lib/domain_layer/usecases/cashu/cashu.dart
1-leo: fixed with `_ensureMintCounterSafety()`

packages/ndk/lib/domain_layer/usecases/cashu/cashu.dart
1-leo: is now guarded with a mutex.

frnandu (approved):
utACK

nogringo (approved):

PR #827, opened 2026-09-15
feat: sample app/cashu restore quote

### LINKED ISSUE #809: Cashu: mint quote lock key is random and unrecoverable; paid quotes are abandoned when the app closes

1-leo:
Flutter ui for `restore ecash quote` stacked on top of https://github.com/relaystr/ndk/pull/818


frnandu (approved):

packages/ndk_flutter/lib/widgets/wallets/cashu_restore.dart
```diff
@@ -0,0 +1,271 @@
...
+          TextField(
+            controller: _mintUrlController,
+            decoration: InputDecoration(
+              border: const OutlineInputBorder(),
+              labelText: l10n.mintUrl,
+              hintText: l10n.mintUrlHint,
+              suffixIcon: IconButton(
+                onPressed: _pasteMintUrl,
+                icon: const Icon(Icons.paste),
+                tooltip: l10n.copy,
```
nogringo, line 136: The tooltip on this paste button says "Copy". A `paste` key already exists in `app_en.arb` (`"paste": "Paste"`, exposed as `l10n.paste`), so this can just be `tooltip: l10n.paste,`

Same thing in cashu_quote_recovery.dart:173, on the quote ID field's paste button.

packages/ndk_flutter/lib/widgets/wallets/cashu_restore.dart
```diff
@@ -0,0 +1,271 @@
...
+          ndkFlutter: widget.ndkFlutter,
+          mintUrl: mintUrl,
+        ),
+      );
+
+      if (result != null) {
+        scaffoldMessenger.showSnackBar(
+          SnackBar(
+            content: Text(l10n.restoredFromMint(result.totalProofsRestored)),
+            backgroundColor: Colors.green,
```
nogringo, line 90: When the scan completes without finding anything, this shows a green "Restored 0 proofs from the mint". A scan that found nothing isn't a success, and the green treatment reads as one.

Worth splitting the zero case out with its own string (something like `cashuRestoreNothingFound`: "No proofs found on this mint") and dropping the green background for it:

Note that restoredFromMint isn't plural-aware either, so a single proof reads "Restored 1 proofs". If you're touching the string anyway, an ICU plural would fix both at once.

packages/ndk_flutter/lib/widgets/wallets/cashu_restore.dart
```diff
@@ -0,0 +1,271 @@
...
+            final result = snapshot.data ?? _result;
+            // keyset-level results are cumulative, so keep the last one around
+            // for the summary even after the stream completed
+            if (result != null) _result = result;
+            final bool errored = snapshot.hasError;
+            if (errored) _failed = true;
+            final bool done =
+                !errored && snapshot.connectionState == ConnectionState.done;
+
+            if (done) _finish(result);
```
nogringo, line 216: **A restore where every keyset fails closes silently.**

`restoreAllKeysets` swallows per-keyset exceptions and only logs them (`cashu_restore.dart:383-385` in `packages/ndk`, not touched by this PR), and it yields *inside* the `try`. So when every keyset throws, the stream completes having emitted neither data nor an error.

Following that through this dialog:

- `errored` is false, so `done` is true and `_finish(result)` fires with `result == null`
- back in `_startRestore`, the `if (result != null)` guard at line 86 skips the snackbar, then `Navigator.pop(result)` closes the input dialog too

Both dialogs disappear with no message at all, and the user reasonably concludes the restore worked and found nothing. Contrast with a mint that's simply unreachable: `getKeysetsFromMint` throws before the loop, so that case *does* surface an error panel. It's specifically the all-keysets-failed path that goes quiet.

packages/ndk_flutter/lib/widgets/wallets/cashu_quote_recovery.dart
```diff
@@ -0,0 +1,344 @@
...
+  Widget build(BuildContext context) {
+    final l10n = AppLocalizations.of(context)!;
+    return AlertDialog(
+      title: Text(l10n.cashuQuoteRecoveryTitle),
+      content: SizedBox(
+        width: double.maxFinite,
+        child: StreamBuilder<CashuQuoteRecoveryProgress>(
+          stream: _progress,
+          builder: (context, snapshot) {
+            if (snapshot.hasError && _errorMessage == null) {
```
nogringo, line 252: `_errorMessage` is write-only. It's assigned here and never read anywhere else: `_failureMessage` below re-derives the message from `snapshot.hasError` / `snapshot.error` directly, and the panel is fed from that. So this block and the field at line 230 can both go.

packages/ndk_flutter/lib/l10n/app_en.arb
```diff
@@ -1694,89 +1497,122 @@
...
+  },
+  "quoteIdHint": "Paste the quote ID here",
+  "@quoteIdHint": {
+    "description": "Hint text for the quote ID input field"
+  },
+  "pleaseEnterQuoteId": "Please enter a quote ID",
+  "@pleaseEnterQuoteId": {
+    "description": "Validation message when the quote ID field is empty"
+  },
+  "recoveringQuote": "Recovering quote...",
```
nogringo, line 1794: `recoveringQuote` is never used. Its description says "Button label while a quote recovery is in progress", but the recover button renders a `CircularProgressIndicator` instead of a label while `_starting` is true.

nogringo (changes requested):
`app_en.arb` gained 24 keys, but none of the 11 translated arb files (`app_de.arb`, `app_fr.arb`, ...) were updated, so `gen_l10n` fell back to the template when generating the locale classes.

Three UX notes:

**Two entries called "Restore".** `n_wallet_card.dart:564` is the backup-JSON
import; the new one at line 586 is the seed-based mint scan. Renaming the
existing label to "Restore from backup" would separate them.

**"Close" cancels the recovery.** `retrieveFunds` polls in a `while (true)`
until paid or expired (`cashu.dart:1161`), and closing the dialog
(`cashu_quote_recovery.dart:310`) disposes the `StreamBuilder`, which kills the
generator. Nothing is lost (the transaction stays `pending` and auto-resume
picks it up), but the label reads as "hide", not "stop waiting".

**Failures show twice:** the `CashuErrorPanel` at line 299, then the same detail
as a red snackbar at lines 109-118. I'd keep the panel and drop the snackbar for
the failure case.

frnandu:
if it is stacked on top, shouldn't this PR be to merge into `1-leo/cashu-mint-quote-lock-key-is-random-and-unrecove` instead of master?
otherwise it will unsync missing commits from `1-leo/cashu-mint-quote-lock-key-is-random-and-unrecove`...

frnandu:
after addressing the review feedback from @nogringo we should close it in favor of #831 which is properly stacked on top of #818

frnandu:
I've copied @nogringo comments into new PR #831 which replaces this one

PR #828, opened 2026-09-16
feat: mod popup, sample app demo

1-leo:
Message of the Day feature.

As a developer I want to directly inform users about an incompatible version or other actionable circumstances.

On purpose only implemented in `ndk_flutter` not a `ndk` usecase.

The data is kind 30078 nostr event with `mod` d tag.
App developers can hardcode their nostr acc as author so the widget can query them

```dart
NMotdController({
 required NdkFlutter ndkFlutter,
 required String authorPubkey,
 String dTagValue = MotdData.kDefaultDTag,
 Duration queryTimeout = const Duration(seconds: 15),
})

Declared in NMotdController in package:ndk_flutter/widgets/message_of_the_day/n_motd_controller.dart.

```

image

image

## Summary by CodeRabbit

* **New Features**
  * Added Message of the Day popups with configurable styling, titles, messages, links, and dismissal behavior.
  * Added controls for authorized users to publish, update, and delete messages.
  * Added minimum app-version support to target messages to compatible app versions.
  * Integrated Message of the Day functionality into the sample app.
* **Localization**
  * Added Message of the Day interface text and validation messages across supported languages.
* **Tests**
  * Added coverage for popup display, version filtering, dismissal, publishing, editing, deletion, and authorization.


packages/ndk_flutter/lib/widgets/message_of_the_day/n_motd_controller.dart
```diff
@@ -0,0 +1,177 @@
...
+    this.dTagValue = MotdData.kDefaultDTag,
+    this.queryTimeout = const Duration(seconds: 15),
+  });
+
+  /// Current lookup status.
+  NMotdStatus get status => _status;
+
+  /// The fetched message, or null when none exists (yet).
+  MotdData? get current => _current;
+
```
frnandu, line 60: 🔴 bug: `Minimum version` logic reversed; event `1.2.0` shows on app `1.0.0`, hides on `1.2.0+`. Compare `appVersion` against minimum, hide only when app < minimum.

packages/ndk_flutter/lib/widgets/message_of_the_day/n_motd_admin.dart
```diff
@@ -0,0 +1,349 @@
...
+    final url = _urlController.text.trim();
+    final version = _versionController.text.trim();
+
+    setState(() {
+      _status = _AdminStatus.saving;
+      _error = null;
+    });
+
+    try {
+      final event = Nip01Event(
```
frnandu, line 147: 🟡 risk: two updates within same second share replaceable-event timestamp; relays can retain old content. Ensure next `createdAt` exceeds existing event timestamp.

frnandu (changes requested):
Changes requested for version targeting and replaceable-event ordering.

frnandu:
clicking show popup after having published does nothing...
does it depend on which account pubkey does the publishing??

1-leo:
@frnandu that is intended there is a remember function to only show once. So it does not annoy the user on every launch.
Can make this configurabale or with a  backoff time, wdyt?

frnandu:
Oh, I thought that for testing in sample app clicking would always open, specially for admin edition for checking if it look ok before publishing.

packages/ndk_flutter/lib/widgets/message_of_the_day/n_motd_controller.dart
```diff
@@ -0,0 +1,177 @@
...
+          dTags: [dTagValue],
+          limit: 1,
+        ),
+        timeout: queryTimeout,
+      );
+      final events = await response.future;
+
+      Nip01Event? latest;
+      for (final event in events) {
+        if (event.kind == MotdData.kKind &&
```
frnandu, line 115: **Security — enforce the configured author locally.** This accepts any event with the expected kind and `d` tag, without checking `event.pubKey == authorPubkey`. NDK verifies the signature, but its relay receive path forwards events without checking them against the requested author filter. A malicious relay can therefore return a valid event signed by an attacker and make the app display attacker-controlled MOTD text and link. Check the author here and in `NMotdAdmin._loadExisting()`; ideally also enforce request filters before NDK caches relay events. Add a wrong-author event test.

packages/ndk_flutter/lib/widgets/message_of_the_day/n_motd_popup.dart
```diff
@@ -0,0 +1,188 @@
...
+      context: context,
+      barrierDismissible: widget.config.barrierDismissible,
+      builder: (dialogContext) => _NMotdDialog(
+        motd: motd,
+        config: widget.config,
+        onClose: () => Navigator.of(dialogContext).pop(),
+        onOpenLink: () async {
+          final url = motd.url;
+          if (url == null) return;
+          final launched = await launchUrl(
```
frnandu, line 86: **Security — restrict event-provided links to approved schemes.** `Uri.parse(url)` is passed directly to `launchUrl(..., mode: LaunchMode.externalApplication)`. A forged MOTD from the author-check issue, or a compromised publisher key, can place a native deep link or other non-web URI behind the generic “Learn more” button. Parse and validate the URI before rendering/enabling the button; accept `https` only unless the host app explicitly opts into other schemes. Reject malformed URIs without throwing from the button callback.

packages/ndk_flutter/lib/widgets/message_of_the_day/n_motd_controller.dart
```diff
@@ -0,0 +1,177 @@
...
+      final response = ndkFlutter.ndk.requests.query(
+        filter: Filter(
+          authors: [authorPubkey],
+          kinds: [MotdData.kKind],
+          dTags: [dTagValue],
+          limit: 1,
+        ),
+        timeout: queryTimeout,
+      );
+      final events = await response.future;
```
frnandu, line 111: **Performance — bound results from noncompliant relays.** `response.future` collects the entire event stream into a list. `Filter(limit: 1)` is sent to relays, not enforced on this response, so a relay can flood events until the 15-second timeout; NDK verifies and caches them before this code selects one. Bound accepted events or bytes locally and stop/cancel the query once that bound is reached. The admin query uses the same `response.future` pattern and needs the same bound.

frnandu (changes requested):
Request changes for MOTD event trust boundary, external URL handling, and relay-driven query work. Please address inline findings before merge.

PR #831, opened 2026-09-16
feat: restore quote ui

frnandu:
## Summary by CodeRabbit

* **New Features**
  * Added Cashu quote recovery with mint URL and quote ID entry, clipboard support, validation, live progress, and success or failure feedback.
  * Added restoration of funds from a Cashu mint, including keyset scanning, proof recovery, and pending quote recovery.
  * Added a centralized Cashu restore and backup menu, including backup restoration and seed phrase management.
  * Added seed phrase validation, override confirmation, and update success or failure feedback.
  * Improved error messages and recovery-stage progress indicators.

* **Localization**
  * Added localized labels, prompts, progress messages, validation text, and results across supported languages.


frnandu:
Originally authored by @nogringo on #827. Copied here because #831 replaces that PR.

[Original inline comment](https://github.com/relaystr/ndk/pull/827#discussion_r4023888066)

Original context: `packages/ndk_flutter/lib/widgets/wallets/cashu_restore.dart:136`.

---

The tooltip on this paste button says "Copy". A `paste` key already exists in `app_en.arb` (`"paste": "Paste"`, exposed as `l10n.paste`), so this can just be `tooltip: l10n.paste,`

Same thing in cashu_quote_recovery.dart:173, on the quote ID field's paste button.

frnandu:
Originally authored by @nogringo on #827. Copied here because #831 replaces that PR.

[Original inline comment](https://github.com/relaystr/ndk/pull/827#discussion_r4023921072)

Original context: `packages/ndk_flutter/lib/widgets/wallets/cashu_restore.dart:90`.

---

When the scan completes without finding anything, this shows a green "Restored 0 proofs from the mint". A scan that found nothing isn't a success, and the green treatment reads as one.

Worth splitting the zero case out with its own string (something like `cashuRestoreNothingFound`: "No proofs found on this mint") and dropping the green background for it:

Note that restoredFromMint isn't plural-aware either, so a single proof reads "Restored 1 proofs". If you're touching the string anyway, an ICU plural would fix both at once.

frnandu:
Originally authored by @nogringo on #827. Copied here because #831 replaces that PR.

[Original inline comment](https://github.com/relaystr/ndk/pull/827#discussion_r4023959536)

Original context: `packages/ndk_flutter/lib/widgets/wallets/cashu_restore.dart:216`.

---

**A restore where every keyset fails closes silently.**

`restoreAllKeysets` swallows per-keyset exceptions and only logs them (`cashu_restore.dart:383-385` in `packages/ndk`, not touched by this PR), and it yields *inside* the `try`. So when every keyset throws, the stream completes having emitted neither data nor an error.

Following that through this dialog:

- `errored` is false, so `done` is true and `_finish(result)` fires with `result == null`
- back in `_startRestore`, the `if (result != null)` guard at line 86 skips the snackbar, then `Navigator.pop(result)` closes the input dialog too

Both dialogs disappear with no message at all, and the user reasonably concludes the restore worked and found nothing. Contrast with a mint that's simply unreachable: `getKeysetsFromMint` throws before the loop, so that case *does* surface an error panel. It's specifically the all-keysets-failed path that goes quiet.

frnandu:
Originally authored by @nogringo on #827. Copied here because #831 replaces that PR.

[Original inline comment](https://github.com/relaystr/ndk/pull/827#discussion_r4023982067)

Original context: `packages/ndk_flutter/lib/widgets/wallets/cashu_quote_recovery.dart:252`.

---

`_errorMessage` is write-only. It's assigned here and never read anywhere else: `_failureMessage` below re-derives the message from `snapshot.hasError` / `snapshot.error` directly, and the panel is fed from that. So this block and the field at line 230 can both go.

frnandu:
Originally authored by @nogringo on #827. Copied here because #831 replaces that PR.

[Original inline comment](https://github.com/relaystr/ndk/pull/827#discussion_r4024044644)

Original context: `packages/ndk_flutter/lib/l10n/app_en.arb:1524`.

---

`recoveringQuote` is never used. Its description says "Button label while a quote recovery is in progress", but the recover button renders a `CircularProgressIndicator` instead of a label while `_starting` is true.

frnandu:
Originally authored by @nogringo on #827. Copied here because #831 replaces that PR.

[Original review](https://github.com/relaystr/ndk/pull/827#pullrequestreview-5220189264)

---

`app_en.arb` gained 24 keys, but none of the 11 translated arb files (`app_de.arb`, `app_fr.arb`, ...) were updated, so `gen_l10n` fell back to the template when generating the locale classes.

Three UX notes:

**Two entries called "Restore".** `n_wallet_card.dart:564` is the backup-JSON
import; the new one at line 586 is the seed-based mint scan. Renaming the
existing label to "Restore from backup" would separate them.

**"Close" cancels the recovery.** `retrieveFunds` polls in a `while (true)`
until paid or expired (`cashu.dart:1161`), and closing the dialog
(`cashu_quote_recovery.dart:310`) disposes the `StreamBuilder`, which kills the
generator. Nothing is lost (the transaction stays `pending` and auto-resume
picks it up), but the label reads as "hide", not "stop waiting".

**Failures show twice:** the `CashuErrorPanel` at line 299, then the same detail
as a red snackbar at lines 109-118. I'd keep the panel and drop the snackbar for
the failure case.

frnandu:
maybe all the restore/recover/reclaim options should be inside one menu option `recover` and in dialog, so that menu doesn't explode in complexity with so many recover/restore advanced options that normal user shouldn't need to care about

this menu is way too complicated at first glance, it will only scare away from using cashu things IMO:

Screenshot from 2026-09-16 23-58-11

1-leo:
Fixing:
- @frnandu: comment — Originally authored by @nogringo on #827. Copied here because #831 repl…
- @frnandu: comment — Originally authored by @nogringo on #827. Copied here because #831 repl…
- @frnandu: comment — Originally authored by @nogringo on #827. Copied here because #831 repl…

Will be in the next commit.

nogringo (changes requested):
Two blocking issues in `cashu_seed_phrase.dart`, both on the save path.

**1. Validation and persistence use different strings**

`_validateSeedPhrase` validates `words.join(' ')`, so whitespace is normalized, but `_save` applies and persists the raw `phrase`. Since the field is `maxLines: 4` and `Mnemonic.fromSentence` splits on a single space, a seed pasted across several lines passes validation and then fails inside `CashuSeed`, after the corrupt phrase has already been written to secure storage.

Repro against the real package: validation passes, then `setSeedPhrase` throws `Expect mnemonic [12, 15, 18, 21, 24] words, got 11`, and `isSeedPhraseSet` stays `false`.

Fix: normalize once and use that same string in both places, e.g.

```dart
final normalized = phrase.trim().split(RegExp(r'\s+')).join(' ');
```

**2. `setCashuSeedPhrase` is not awaited**

`Cashu.setCashuSeedPhrase` is a `void` wrapper over the `async` `CashuSeed.setSeedPhrase`, so a rejection surfaces as an unhandled async error instead of reaching the dialog. The `try`/`catch` in `_save` never fires: the dialog pops `true` and shows the green "Seed phrase updated" snackbar while Cashu has no usable seed.

Fix: make the wrapper return `Future<void>` and `await` it in `_save`, so the existing error handling actually runs.

Together these mean an invalid seed can be stored and reported as a success. Worth fixing before merge.

packages/ndk/lib/domain_layer/usecases/cashu/cashu.dart
```diff
@@ -153,9 +153,10 @@ class Cashu {
 
   /// set cashu user seed phrase, required for using cashu features \
   /// ideally use the NdkConfig to set the seed phrase on initialization \
-  /// you can use CashuSeed.generateSeedPhrase() to generate a new seed phrase
-  void setCashuSeedPhrase(CashuUserSeedphrase userSeedPhrase) {
-    _cashuSeed.setSeedPhrase(seedPhrase: userSeedPhrase.seedPhrase);
+  /// you can use CashuSeed.generateSeedPhrase() to generate a new seed phrase \
+  /// throws if the seed phrase is invalid
+  Future<void> setCashuSeedPhrase(CashuUserSeedphrase userSeedPhrase) async {
```
nogringo, line 158: Only `seedPhrase` is forwarded, even though `CashuUserSeedphrase` also carries `language` and `passphrase`, and `CashuSeed.setSeedPhrase` accepts both. The init path does this correctly (`cashu_seed.dart:24-28`), so the two paths diverge: re-setting the very same `CashuUserSeedphrase` that was passed at init derives a different seed as soon as the language isn't English or a passphrase is in use, and existing proofs become unspendable.

The bug predates this PR, but this PR is what makes the path reachable from the UI through the new dialog, and the line is already being modified here, so it seems worth fixing along the way.

packages/ndk_flutter/lib/widgets/wallets/cashu_seed_phrase.dart
```diff
@@ -0,0 +1,232 @@
...
+      _error = null;
+    });
+    try {
+      // Apply on the running instance, then persist for future launches.
+      // Awaiting the actual seed derivation means a failure surfaces here
+      // instead of as an unhandled async error.
+      await widget.ndkFlutter.ndk.cashu.setCashuSeedPhrase(
+        CashuUserSeedphrase(seedPhrase: phrase),
+      );
+      await widget.seedStore.write(phrase);
```
nogringo, line 108: If `write` fails, the state is inconsistent: the `catch` shows `cashuSeedPhraseUpdateFailed` and the user believes nothing changed, but the in-memory `Cashu` is already using the new seed. Anything minted afterwards is derived from a seed that exists nowhere, and is gone on restart since secure storage still holds the old one. Swapping the order is enough: persist first, apply second.

In that order, a failed `write` leaves the app untouched on the old seed, which is a recoverable state. The validation done earlier in `_save` already covers most of the risk of `setCashuSeedPhrase` failing after a successful write.

nogringo (changes requested):
Two issues on the seed phrase path. Both lead to unspendable proofs.

